PR #17, opened 2021-12-06
WIP: Fixes for aws merge branch

laestrada:
@msulprizio @djvaron. Still a few things to work out, specifically the dry run simulations are looking for data that doesn't exist on gcgrid. I have left a few comments on fixes that I am not sure what the consequences are.


setup_ch4_inversion.sh
```diff
@@ -202,7 +203,7 @@ if "$CreateStateVectorFile"; then
     printf "\n=== CREATING STATE VECTOR FILE ===\n"
     
     # Use GEOS-FP or MERRA-2 CN file to determine ocean/land grid boxes
-    LandCoverFile="${DataPath}/GEOS_${gridDir}/${metDir}/${constYr}/01/${metUC}.${constYr}0101.CN.${gridRes}.nc"
+    LandCoverFile="${DataPath}/GEOS_${gridDir}_NA/${metDir}/${constYr}/01/${metUC}.${constYr}0101.CN.${gridRes}.NA.nc"
```
laestrada, line 206: The GEOS_0.25x0.3125_NA landcover files for 2018 is already present on the ami, but the GEOS_0.25x0.3125 data for 2018 is not present -- there is only 2011 data. My assumption is that we should use the NA data, but I don't know the difference? Or if we should download the corresponding landcover files from elsewhere?

setup_ch4_inversion.sh
```diff
@@ -463,8 +468,8 @@ if  "$SetupSpinupRun"; then
     sed -i -e "s|${StartDate}|${SpinupStart}|g" \
            -e "s|${EndDate}|${SpinupEnd}|g" \
            -e "s|Do analytical inversion?: T|Do analytical inversion?: F|g" \
-           -e "s|{PERTURBATION}|1.0|g" \
-           -e "s|{ELEMENT}|0|g" input.geos
+           -e "s|pertpert|1.0|g" \
+           -e "s|clustnumclustnum|0|g" input.geos
```
laestrada, line 472: Currently {PERTURBATION} and {ELEMENT} are not present in the template input.geos.CH4 file, so I reverted this back to the original sed command. 

However, it seems like we should probably change the input.geos.CH4 file to have {PERTURBATION} and {ELEMENT} instead of pertpert and clustnumclustnum, but this would require updating the GCClassic code and updating to a version with that update. Is this correct @msulprizio?

setup_ch4_inversion.sh
djvaron: I'm also not sure about this. Would seem OK to me to use a land-cover file for 2011, and we probably want one that can be used for any domain. Not sure if we need a different file for each year, or just select one to use as default? Using a global file for a particular year as default would be one less thing to worry about.

laestrada: Okay, then I will update this to use the 2011 land cover file instead👍

msulprizio: Do we know why it's looking for a landcover file for 2018? It should only look for the constant year (`constYr`) for that met field since landcover is a constant field. For GEOS-FP, `constYr=2011` and for MERRA-2 `constYr=2015` -- both values are defined further up in `setup_ch4_inversion.sh`.

msulprizio: Also, we should use the NA file by default since our example case uses a domain in North America (and NA met fields). If users choose a different domain like `CH`, `AS`, or `EU` then we should have the script automatically replace the region token in that path (I can look into that). Users may also choose to manually update this file to use a different region if they've pre-cropped met fields on their end. The global, high-resolution land cover file may also be used -- we just avoid using those files for all met fields because of the negative impact in I/O when reading/regridding those fields in HEMCO.

djvaron: Right, and `${gridDir}` include `"_NA"` if `$Region="NA"` as by default. Therefore we shouldn't need to specify `GEOS_${gridDir}_NA`

setup_ch4_inversion.sh
msulprizio: I have those locally but have not pushed those changes to our GEOS-Chem fork. I will do so ASAP.

setup_ch4_inversion.sh
djvaron: > Also, we should use the NA file by default since our example case uses a domain in North America (and NA met fields). If users choose a different domain like `CH`, `AS`, or `EU` then we should have the script automatically replace the region token in that path (I can look into that). Users may also choose to manually update this file to use a different region if they've pre-cropped met fields on their end. The global, high-resolution land cover file may also be used -- we just avoid using those files for all met fields because of the negative impact in I/O when reading/regridding those fields in HEMCO.

Right, I haven't been able to figure out why the region token isn't being included automatically, as I alluded to above. Also, I didn't think about region-specific land-cover files. But that's totally fine, as long as we have one to go along with the met data for each of the `NA`, `CH`, `AS`, `EU` domains.

setup_ch4_inversion.sh
laestrada: Great, thanks!

setup_ch4_inversion.sh
laestrada: I don't know why the Region wasn't getting picked up. I will rerun the setup script and see what is going on.

PR #48, opened 2022-05-02
Cleanup IMI configuration file, setup script, and run script

msulprizio:
This PR includes:

1. Moving several fields from hard-coded values in `setup_imi.sh` and `run_imi.sh` to configurable fields in `config.yml`.
2. Adding the capability to specify a configuration file when executing `setup_imi.sh` or `run_imi.sh`. If no configuration file is specified, `config.yml` will be used.
3. Adding a sample configuration file for running the IMI on a local cluster (in this case Harvard's Cannon cluster).
4. Adding copies of GEOS-Chem run scripts to the IMI repository instead of copying these files from a fork of GEOS-Chem.


config.yml
```diff
@@ -75,10 +78,63 @@ DoPosterior: false
...
+##------------------------------------------------------------------
+
+## Path for IMI runs and output
+OutputPath: "/home/ubuntu/imi_output_dir"
+
+## Path to GEOS-Chem input data
+DataPath: "/home/ubuntu/ExtData"
+
+## Environment files
+source /home/ubuntu/miniconda/etc/profile.d/conda.sh
```
laestrada, line 120: @msulprizio can yaml files run `source`? Or was this intended to just include the conda path as a yaml variable?

config.yml
```diff
@@ -75,10 +78,63 @@ DoPosterior: false
...
+CondaEnv: "geo"
+
+## Download initial restart file from AWS S3?
+RestartDownload: true
+
+## Path to initial GEOS-Chem restart file
+RestartFile: "${DataPath}/BoundaryConditions/GEOSChem.BoundaryConditions.${SpinupStart}_0000z.nc4"
+RestartFilePreview: "${DataPath}/BoundaryConditions/GEOSChem.BoundaryConditions.${StartDate}_0000z.nc4"
```
laestrada, line 128: Do bash variables get resolved by the `parse_yaml` function? eg. `${DataPath}`

laestrada (commented):
Just a few questions!

config.yml
msulprizio: That's a typo. I will fix it! Thanks for catching that.

config.yml
msulprizio: Good point. I can update that to include the full path.

PR #56, opened 2022-08-29
feature/area weighting for averaging obs

laestrada:


djvaron:
I think this looks good. Just to make sure I understand conceptually:

- Each GEOS-Chem grid cell has its own `gridcell_dict`
- TROPOMI observations that fall within a single GC grid cell (which is most observations) get put into the corresponding `gridcell_dict` with an observation weight of 1
- Observations that intersect multiple GC grid cells get put into multiple `gridcell_dict` objects, each with a fractional observation weight depending on the overlap area fraction
- Once all the observations are sorted into one or more `gridcell_dict` objects, the weighted averaging is performed for each `gridcell_dict`
- Observations that fall within a single grid cell get unit weight in that grid cell, and observations that fall within multiple grid cells get fractional weight in multiple grid cells. 

Is that interpretation correct? If so, it's kind of inverting the way we used to do the weighted average (weights were applied to GC values to compare with TROPOMI, whereas now it's the opposite) -- but I think it makes sense.

Specific comments:
- Line 469: is it really the grid cell with the most overlap area, or is that phrasing a relic from a previous commit? I think you're looping through the overlaps now as opposed to selecting the max
- Line 473: by the same token, shouldn't this be `gridcell_dict["overlap_area"].append(overlap)`? i.e., save the overlap for the current loop index, rather than always the max overlap?

laestrada:
> I think this looks good. Just to make sure I understand conceptually:
> 
> * Each GEOS-Chem grid cell has its own `gridcell_dict`
> * TROPOMI observations that fall within a single GC grid cell (which is most observations) get put into the corresponding `gridcell_dict` with an observation weight of 1
> * Observations that intersect multiple GC grid cells get put into multiple `gridcell_dict` objects, each with a fractional observation weight depending on the overlap area fraction
> * Once all the observations are sorted into one or more `gridcell_dict` objects, the weighted averaging is performed for each `gridcell_dict`
> * Observations that fall within a single grid cell get unit weight in that grid cell, and observations that fall within multiple grid cells get fractional weight in multiple grid cells.
> 
> Is that interpretation correct? If so, it's kind of inverting the way we used to do the weighted average (weights were applied to GC values to compare with TROPOMI, whereas now it's the opposite) -- but I think it makes sense.
> 
> Specific comments:
> 
> * Line 469: is it really the grid cell with the most overlap area, or is that phrasing a relic from a previous commit? I think you're looping through the overlaps now as opposed to selecting the max
> * Line 473: by the same token, shouldn't this be `gridcell_dict["overlap_area"].append(overlap)`? i.e., save the overlap for the current loop index, rather than always the max overlap?

@djvaron Your understanding is correct. And yes, I need to clean up some comments and remove some unnecessary code now. Thanks for your feedback!

PR #57, opened 2022-09-01
[PULL REQUEST] Average TROPOMI observations

laestrada:
This pr includes:
- the addition of averaging tropomi retrievals for each GC gridcell
- the addition of a weighting function (provided by zichong) to prevent overfitting
- updating default gamma to 1
- updates to how we area weight observations in multiple gridcells
- I will flag areas for potential comment (most of the changes are in jacobian.py)


src/inversion_scripts/jacobian.py
```diff
@@ -449,173 +689,100 @@ def apply_tropomi_operator(
...
-                    overlap_area[gridcellIndex] * tropomi_sensitivity_gridcellIndex
-                )  # m2
+            area_weighted_virtual_tropomi_sensitivity += (
+                gridcell_dict["overlap_area"] * tropomi_sensitivity_gridcellIndex
+            )  # m2
 
         # Compute virtual TROPOMI observation as weighted mean by overlapping area
         # i.e., need to divide out area [m2] from the previous step
-        virtual_tropomi = area_weighted_virtual_tropomi / sum(overlap_area)
+        virtual_tropomi = area_weighted_virtual_tropomi / gridcell_dict["overlap_area"]
```
laestrada, line 767: I am not sure exactly whether this is still necessary... the overlap area is now the averaged overlap area...

djvaron: I don't think we need any of these area-weighting lines that multiply or divide by overlap area -- because you already dealt with the areas when building `gridcell_dicts`, right? Now you just have one average TROPOMI observation for each GEOS-Chem grid cell. So you just need to map each grid cell to a TROPOMI once (rather than doing it for each individual observation, including ones that overlap with multiple grid cells). It looks like this code is multiplying and then dividing by `gridcell_dict["overlap_area"]` in a couple of places.

I also think we should review the comments here to clarify. E.g., in places above we say "For each TROPOMI observation..." but what we're really talking about now are grid-cell-averaged TROPOMI observations.

src/inversion_scripts/jacobian.py
```diff
@@ -449,173 +689,100 @@ def apply_tropomi_operator(
...
+        # Derive the column-averaged XCH4 that TROPOMI would see over this ground cell
+        # using eq. 46 from TROPOMI Methane ATBD, Hasekamp et al. 2019
+        virtual_tropomi_gridcellIndex = (
+            sum(apriori + avkern * (sat_CH4_molm2 - apriori))
+            / sum(dry_air_subcolumns)
+            * 1e9
+        )  # ppb
+        # Weight by overlapping area (to be divided out later) and add to sum
+        area_weighted_virtual_tropomi += (
+            gridcell_dict["overlap_area"] * virtual_tropomi_gridcellIndex
```
djvaron, line 732: Re: my comment below, we don't need this line, for example, since we already did the area weighting in constructing `gridcell_dicts` and further down end up dividing out `gridcell_dict["overlap_area"]`

src/inversion_scripts/jacobian.py
```diff
@@ -449,173 +689,100 @@ def apply_tropomi_operator(
...
-        if sum(overlap_area) == 0:
-            continue
-
-        # =======================================================
-        #       Map GEOS-Chem to TROPOMI observation space
-        # =======================================================
-
         # Otherwise, initialize tropomi virtual xch4 and virtual sensitivity as zero
         area_weighted_virtual_tropomi = 0  # virtual tropomi xch4
         area_weighted_virtual_tropomi_sensitivity = 0  # virtual tropomi sensitivity
```
djvaron, line 704: I don't think we need these lines anymore either.

PR #84, opened 2022-11-04
updates from sarah for calc_sensi

laestrada:
We should evaluate whether these changes should go into the standard IMI code. Sarah writes:
"An important thing to note though is that while it speeds things up, it does require more memory so maybe would depend on the use case if one wants to use my version or the current version. I think there are a lot of things that are different for the large domain case…Like memory default settings, etc."

### DIFF
--- a/src/inversion_scripts/calc_sensi.py
+++ b/src/inversion_scripts/calc_sensi.py
@@ -1,6 +1,8 @@
 import numpy as np
 import xarray as xr
 import datetime
+import gc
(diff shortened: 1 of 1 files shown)


msulprizio:
@laestrada @sarahhancock Can this PR be closed? I imagine it's quite outdated at this point. If these updates are still needed, would it be best to create a new PR?

laestrada:
This will be addressed by James PR speeding up calc_sensi

PR #86, opened 2022-12-05
feature/kalman-filter

djvaron:
First working draft of the Kalman filter capability.

Key updates:
- new `run_kf.sh` run script
- new scripts in `src/kf_scripts/`
- `imi_preview.py` behaves differently if `KalmanMode=true`
- the usual `inversion` directory is now an `inversion_template` directory that gets replicated for individual inversion periods in `kf_inversions/period{i}`


src/kf_scripts/make_periods_csv.py
```diff
@@ -0,0 +1,52 @@
+import datetime
+import pandas as pd
+import sys
+import os
+
+
+def make_periods_csv(first_day, last_day, stepsize_days, save_dir):
```
laestrada, line 7: currently it looks like if you input a time period that is not divisible by `stepsize_days` it will cut off the partial period and not include it in the csv? It might be better to print an error asking the user to either increase or decrease the time period? Or just print a warning that it is not including the remainder?

src/kf_scripts/change_dates.py
```diff
@@ -0,0 +1,59 @@
+import os
```
laestrada, line 1: Is there a reason that we aren't reading the entire yaml file in and then writing it out? Using `startswith` seems a bit error prone?

src/kf_scripts/prepare_sf.py
```diff
@@ -0,0 +1,160 @@
+import xarray as xr
+import os
+import numpy as np
+import yaml
+
+
+def sum_total_emissions(emissions, areas, mask):
```
laestrada, line 7: I think this should be imported instead of copied

src/kf_scripts/multiply_posteriors.py
```diff
@@ -0,0 +1,51 @@
+import xarray as xr
+import os
+
+
+def multiply_posteriors(period_number, base_directory):
```
laestrada, line 5: I wonder if we should add an additional period dimension to the dynamic scale factor file so that we can have a record of the scale factors used at each stage. Maybe this is already being done?

run_kf.sh
```diff
@@ -0,0 +1,408 @@
...
+    
+        cd ${InversionPath}
+
+        # Delete unneeded daily restart files from Jacobian and posterior directories
+        # python ${InversionPath}/src/kf_scripts/cull_restarts.py $JacobianRunsDir $PosteriorRunDir $StartDate_i $EndDate_i
+
+        # Move to next time step
+        echo -e "Moving to next iteration\n"
+
+    done
```
laestrada, line 404: it might be nice to have the timers print out for each iteration

src/kf_scripts/cull_restarts.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import datetime
+import os
+
+
+def cull_restarts(jacobian_runs_dir, posterior_dir, start_day, end_day):
+    """
+    GEOS-Chem bugs mean we need to output daily restart files instead of just one
+    at the end of each week. This function culls the extra restarts we don't need.
+
```
laestrada, line 10: @msulprizio Is this right? Can't we set the restart output frequency to End?

laestrada (changes requested):
Overall looks good! I added a few comment while looking at files. I think that the reorganization of the imi code into components in #88 will make it easier to merge run_imi and run_kf. You should be able to make a src/components/kf_component folder to hold the relevant scripts and python files.

config.yml
```diff
@@ -7,9 +7,15 @@ isAWS: true
 UseSlurm: true
 SafeMode: true
 
+## Kalman filter options
+KalmanMode: true
+UpdateFreqDays: 7
+NudgeFactor: 0.1
+FirstPeriod: 1
```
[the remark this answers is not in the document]
djvaron: It determines the starting point for the current IMI run, within the user's period of interest. For example, if you want to run 3 monthly inversions for May-June-July 2019 and set the corresponding start/end dates (`20190501` and `20190901`) then choosing `FirstPeriod: 1` will execute all three inversions, starting with May. Setting `FirstPeriod: 2` will produce two inversions, starting with June. The reason for this feature is to facilitate re-running inversions for specific sub-periods of the period of interest (rather than having to run everything).

src/kf_scripts/make_periods_csv.py
djvaron: Good point. I think the warning would be useful and most appropriate. I can imagine a user wanting to do weekly inversions for a few months and not caring about some dangling days at the end of the period, but they should definitely be warned.

src/kf_scripts/cull_restarts.py
djvaron: This might be outdated by 2+ years (or maybe it was always wrong).

src/kf_scripts/prepare_sf.py
djvaron: Agreed!

src/kf_scripts/multiply_posteriors.py
djvaron: After discussion, we decided it's best to retain current behavior of saving prior and posterior scale factors for each period to an `archive_sf` directory.

PR #88, opened 2023-01-09
Feature/clustering options

laestrada:
This PR adds the following clustering functionality to the config file:

```
ReducedDimensionStateVector: true
ClusteringPairs:
  - [1, 15]
  - [2, 24]
ForcedNativeResolutionPixels: 
  - [31.5, -104]
```
In this scenario the clustering algorithm would create 39 clusters in the domain of interest with 24 2-grid cell clusters and 15 native resolution pixels. The `ForcedNativeResolutionPixels` option allows you to specify cells as for native resolution by setting the averaging kernel estimate value to 1.0.

The main file changes for the clustering feature are within the following files:
```
- aggregation.py
- imi_preview.py
- statevector.sh
```

Additionally, this PR reorganizes the IMI directories and shell scripts into discrete functions in an attempt to 'componentize' the IMI source code. There are a number of file changes associated with this.

Initial testing with a 1-week inversion shows the following runtime results:
| Step         | 71 elements (v1.1.0) | 71 elements (clustering branch) | 
| ------------ | -------------------- | ------------------------------- | 
| Setup (s)    | 957                  | 487                             |
| Preview (s)  | 214                  | 121                             | 
| Spinup (s)   | 710                  | 710                             | 
| Jacobian (s) | 1530                 | 1500                            | 
| Inversion (s)| 290                  | 290                             | 
| Posterior (s)| 394                  | 390                             | 
| Total (s)    | 3881                 | 3377                            | 
| Posterior emissions (Tg/y)| 0.344                | 0.344                           |

Notably, this PR also reduces the setup time needed significantly from our v1.1.0 benchmark. I think this may be due to the componentization, which reduces some repetition of code (loading environments, etc) and performance enhancements to the IMI preview.

Comparing the 71 element inversions from v1.1.0 and this branch shows no differences as we would expect.


src/components/statevector_component/aggregation.py
```diff
@@ -0,0 +1,390 @@
...
+        num_buffer_elems    int  : number of buffer elements in the state vector
+    Returns:                tuple: zeroed buffer area and zeroed roi
+    """
+    buffer_threshold = int(clusters.max()) - num_buffer_elems
+    result = (clusters <= buffer_threshold) * clusters
+    buffers = (clusters > buffer_threshold) * clusters
+    return result, buffers
+
+
+def scale_buffer_elements(scale_number, buffer):
```
djvaron, line 63: Not sure I understand what this is doing. Shouldn't the buffer elements be static while the ROI elements get clustered together?

djvaron:
I haven't grasped all of the intricacies of `aggregate.py`, but this pull request looks good to me. I really like the "componentization" update. That will be useful for better implementing the Kalman filter feature.

src/components/statevector_component/aggregation.py
laestrada: The buffer elements need to be relabeled once there are fewer sv elements. For example, if there are originally 63 elements in the DOI and 8 buffer elements labeled 64-71. Once the DOI elements have been clustered into, say, 47 elements, the buffer elements need to be relabeled to 48-55. Maybe I should add an additional comment to the description

PR #96, opened 2023-02-26
conditionally change memory for preview

laestrada:
@nicholasbalasus -- going off of your PR:
Conditionally change the memory allocated by sbatch for the imi_preview based on whether running on aws or local cluster.

### DIFF
--- a/src/components/preview_component/preview.sh
+++ b/src/components/preview_component/preview.sh
@@ -99,9 +99,13 @@ run_preview() {
     # if running end to end script with sbatch then use
     # sbatch to take advantage of multiple cores 
     if "$UseSlurm"; then
-        # set number of cores to run preview with
+        # set number of cores and memory to run preview with
         if "$isAWS"; then
-            sed -i -e "s:#SBATCH -c 8:#SBATCH -c ${cpu_count}:g" ${InversionPath}/src/inversion_scripts/imi_preview.py
+            sed -i -e "s:#SBATCH -c 8:#SBATCH -c ${cpu_count}:g" \
+                   -e "s:#SBATCH --mem:##SBATCH --mem:g" ${InversionPath}/src/inversion_scripts/imi_preview.py
+        else
+            sed -i -e "s:##SBATCH:#SBATCH:g" \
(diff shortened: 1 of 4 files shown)


src/components/preview_component/preview.sh
```diff
@@ -99,9 +99,13 @@ run_preview() {
     # if running end to end script with sbatch then use
     # sbatch to take advantage of multiple cores 
     if "$UseSlurm"; then
-        # set number of cores to run preview with
+        # set number of cores and memory to run preview with
         if "$isAWS"; then
-            sed -i -e "s:#SBATCH -c 8:#SBATCH -c ${cpu_count}:g" ${InversionPath}/src/inversion_scripts/imi_preview.py
+            sed -i -e "s:#SBATCH -c 8:#SBATCH -c ${cpu_count}:g" \
```
djvaron, line 104: I don't see `#SBATCH -c 8` in the header of `imi_preview.py`. There are statements for `-n` and `-N` but not `-c`. Do we need `sed` lines to comment those out, like the line below for `--mem`?

Also, do we need sed to add in the `#SBATCH -c ${cpu_count}` line? Or can it use the `-n` flag instead, like this:

`sed -i -e "s:#SBATCH -n 1:#SBATCH -n ${cpu_count}:g" `

src/components/preview_component/preview.sh
```diff
@@ -99,9 +99,13 @@ run_preview() {
...
+                   -e "s:#SBATCH --mem:##SBATCH --mem:g" ${InversionPath}/src/inversion_scripts/imi_preview.py
+        else
+            sed -i -e "s:##SBATCH:#SBATCH:g" \
```
djvaron, line 107: Is this just in case the `imi_preview.py` script has `##` in the SBATCH header? Currently it just has single `#`.

djvaron (commented):
I made some small comments clarifying the `sed` commands that update the `imi_preview.py` SBATCH header. Otherwise looks good to me!

src/components/preview_component/preview.sh
laestrada: In the above case you can see that we comment out the memory if aws is true. I have the uncomment there in case a user accidentally had the wrong setting causing the batch command to be commented out.

PR #98, opened 2023-03-01
add clustering options to Cannon config file

nicholasbalasus:
Adding a missing component of config.yml to the Cannon version of the file.


laestrada (approved):
Thanks for this!

PR #101, opened 2023-03-08
bug fix to make jacobian.py save viz output data

djvaron:
`jacobian.py` should apply both the super-observation TROPOMI operator and the single-observation TROPOMI operator, and save out both results (former for use in the inversion, latter for use in visualization). Currently the viz data isn't being saved. This PR addresses that bug.


laestrada (approved):
looks good!

PR #107, opened 2023-04-12
Add Boundary conditions scripts to IMI source code

laestrada:
First draft of changes from Nick that take advantage of common functions for reading TROPOMI to simplify boundary condition scripts


laestrada (commented):
@nicholasbalasus overall I think this looks great! And greatly simplifies the boundary condition scripts. Nice job

src/write_BCs/write_BCs.ipynb
```diff
@@ -0,0 +1,203 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "start_time_of_interest = np.datetime64(datetime.datetime.strptime(config[\"startdate\"], \"%Y%m%dT%H%M%S\"))\n",
+    "end_time_of_interest = np.datetime64(datetime.datetime.strptime(config[\"enddate\"], \"%Y%m%dT%H%M%S\"))\n",
+    "\n",
+    "def get_TROPOMI_times(filename):\n",
```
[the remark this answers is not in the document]
nicholasbalasus: This shouldn't happen unless the file naming convention changes. (I'll add a test for it though).

src/inversion_scripts/operators/TROPOMI_operator.py
```diff
@@ -256,15 +256,17 @@ def apply_tropomi_operator(
...
         # Find GEOS-Chem lats & lons closest to the corners of the TROPOMI pixel
         longitude_bounds = TROPOMI["longitude_bounds"][iSat, jSat, :]
         latitude_bounds = TROPOMI["latitude_bounds"][iSat, jSat, :]
         corners_lon_index = []
         corners_lat_index = []
         for l in range(4):
-            iGC = nearest_loc(longitude_bounds[l], GEOSCHEM["lon"])
-            jGC = nearest_loc(latitude_bounds[l], GEOSCHEM["lat"])
+            iGC = nearest_loc(longitude_bounds[l], GEOSCHEM["lon"], tolerance=dlon)
+            jGC = nearest_loc(latitude_bounds[l], GEOSCHEM["lat"], tolerance=dlat)
```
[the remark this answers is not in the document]
nicholasbalasus: If we made this change, the tolerance (which used to default to 0.5) would now be dlon=0.3125 and dlat=0.25 for our typical inversions. I don't know if this will catch everything (e.g., if you have a large TROPOMI pixel, missing GEOS-Chem pixels in-between the TROPOMI pixel corners). Maybe we should just change the tolerance to always 5? Or max(dlon and 1) and max(dlat and 1)? @djvaron 

I will also need to change this in the other TROPOMI operator.

src/write_BCs/write_tropomi_GC_daily_avgs.py
```diff
@@ -0,0 +1,141 @@
+import glob
```
laestrada, line 1: This is so much better :)

src/write_BCs/README.md
```diff
@@ -0,0 +1,11 @@
+## Directions to generate the boundary conditions.
+
+1. Change settings in `config_write_BCs.yml`.
+2. Run run_BCs.sh (`sbatch -p seas_compute -t 4-00:00 --mem 4000 --wrap "bash run_BCs.sh"`).
```
laestrada, line 4: Just curious is there a reason you don't just specify the resources in the header of run_BCs.sh?

src/inversion_scripts/operators/TROPOMI_operator.py
```diff
@@ -584,8 +586,8 @@ def average_tropomi_observations(TROPOMI, gc_lat_lon, sat_ind):
         corners_lat_index = []
 
         for l in range(4):
-            iGC = nearest_loc(longitude_bounds[l], gc_lons)
-            jGC = nearest_loc(latitude_bounds[l], gc_lats)
+            iGC = nearest_loc(longitude_bounds[l], gc_lons, tolerance=max(dlon,0.5))
+            jGC = nearest_loc(latitude_bounds[l], gc_lats, tolerance=max(dlat,0.5))
```
laestrada, line 590: Just checking my understanding. Here we use .5 as default for regular inversion runs, but since we are on a 4x5 grid we need to update this accordingly eg to 4 or 5? Wouldn't we want .25 for .25 inversion runs?

laestrada (commented):
This looks great Nick! Mostly just clarification questions.

src/write_BCs/README.md
nicholasbalasus: No reason, I just get caught up in my --wrap ways.

src/inversion_scripts/operators/TROPOMI_operator.py
nicholasbalasus: This update just allows the TROPOMI operators to be both (1) consistent with previous behavior of the IMI (0.5 degree default) and (2) useful for global simulations like 4 x 5. We wouldn't necessarily want to set this to dlon always. An example is a 1 degree wide TROPOMI pixel (admittedly rare). If we search 0.3125 degrees from its corners for GC grid cell centers, we would miss the one at 0.47.
image

This isn't a perfect fix ([this ](https://github.com/geoschem/integrated_methane_inversion/issues/110) should address it).

PR #108, opened 2023-04-14
Update download_TROPOMI.py to use v02.04.00/v02.05.00 of tropomi data

laestrada:
Ensures download of v2.0.4 of the tropomi data. Additionally, simplifies the logic used to determine the subdirectory to download from. There was previously some convoluted logic here that I attempted to simplify and make more readable.

Expected behavior is that data before 07-26-2022 is downloaded from the `s3://meeo-s5p/RPRO` bucket and everything after is downloaded from `s3://meeo-s5p/OFFL`.

@nicholasbalasus can you take a look since you are the tropomi expert?


src/utilities/download_TROPOMI.py
```diff
@@ -15,186 +18,62 @@ def download_TROPOMI(startdate, enddate, Sat_datadir):
...
+    # This is supposedly faster than downloading data with boto3 via python
     DATA_DOWNLOAD_SCRIPT = "./auto_generated_download_script.sh"
     cmd_prefix = "aws s3 sync --only-show-errors "
     remote_root = "s3://meeo-s5p/"
-    # access number of days in each month easily
-    month_days = [31, [28, 29], 31, 30, 31, 30, 31, 31, 30, 31, 30, 31]
+
     with open(DATA_DOWNLOAD_SCRIPT, "w") as f:
         print("#!/bin/bash\n", file=f)
         print("# This script was generated by jacobian.py\n", file=f)
```
nicholasbalasus, line 38: I'm not sure this is true (this function is never called in jacobian.py, only in run_imi.sh).

src/utilities/download_TROPOMI.py
```diff
@@ -15,186 +18,62 @@ def download_TROPOMI(startdate, enddate, Sat_datadir):
...
-    start_year = start_str[:4]
-    start_month = start_str[5:7]
-    start_day = start_str[8:10]
 
-    end_str = str(enddate)
-    end_year = end_str[:4]
-    end_month = end_str[5:7]
-    end_day = end_str[8:10]
+    # list of all dates between startdate and enddate excluding enddate
+    download_dates = np.arange(startdate, enddate, dtype="datetime64[D]")
```
nicholasbalasus, line 28: I think you might want `download_dates = np.arange(startdate, enddate + np.timedelta64(1, 'D'), dtype="datetime64[D]")`.

laestrada: I think you might be right. I was mimicking the previous script, which excludes the end date, but I think we would want this to be inclusive. @djvaron thoughts?

nicholasbalasus:
Thanks, Lucas, this looks great. It's really cool that they got all of the RPRO data on AWS so fast. The pull request might be better named "Update download_TROPOMI.py" to use v02.04.00/v02.05.00 of TROPOMI data." There is a switch from v02.04.00 to v02.05.00 in March 2023, but it doesn't impact the CH4 retrieval.

nicholasbalasus:
Not super important, but just emphasizing it is v02.04.00/v02.05.00 (instead of v2.0.4 and v2.0.5).

src/utilities/download_TROPOMI.py
djvaron: I'm not so sure. Currently our inversions end at hour 0 of the end date. So if I want a 1-week inversion starting on 2018-05-01, I set `startdate = 20180501` and `enddate = 20180508`. Days 1-7 are complete, and day 8 includes just hour 0. This is probably why we exclude the `enddate` -- right?

PR #109, opened 2023-04-19
Feature/offshore emissions

djvaron:
These updates make it easy to include offshore emissions in the state vector. 

- The state vector script now expects a default global HEMCO diagnostics file from a HEMCO stand-alone run at 0.25 or 0.5 deg resolution (need to add to `gcgrid` on AWS).
- It uses the diagnostic file to include offshore grid elements with emissions above a threshold
- The threshold is defined in config.yml (default 0)
- The state vector component was updated to reflect the changes
- Updates tested on Harvard Cannon but not on AWS


config.yml
```diff
@@ -43,6 +43,7 @@ CreateAutomaticRectilinearStateVectorFile: true
 nBufferClusters: 8
 BufferDeg: 5
 LandThreshold: 0.25
+OffshoreEmisThreshold: 0
```
laestrada, line 46: Can you update the readthedocs page with a corresponding description of this new config variable?

Also, add it to `sanitize_input_yaml.py` since this should be a required variable?

src/components/statevector_component/statevector.sh
```diff
@@ -30,7 +33,7 @@ create_statevector() {
     chmod 755 make_state_vector_file.py
 
     printf "\nCalling make_state_vector_file.py\n"
-    python make_state_vector_file.py $LandCoverFile $StateVectorFName $LatMin $LatMax $LonMin $LonMax $BufferDeg $LandThreshold $nBufferClusters
+    python make_state_vector_file.py $LandCoverFile $HemcoDiagFile $StateVectorFName $LatMin $LatMax $LonMin $LonMax $BufferDeg $LandThreshold $OffshoreEmisThreshold $nBufferClusters
```
laestrada, line 36: The parameters for this python script are getting crowded. Maybe we should just have it take in the config file path as a variable and have the script read the yaml file directly? Just a thought.

src/utilities/make_state_vector_file.py
```diff
@@ -46,49 +46,60 @@ def check_nested_grid_compatibility(lat_min, lat_max, lon_min, lon_max, land_cov
...
+        emis_threshold [float] : Minimum emission to include offshore pixel as a state vector element
         k_buffer_clust [int]   : Number of buffer clusters for k-means
 
     Returns
         ds_statevector []     : xarray dataset containing state vector field formatted for HEMCO
 
     Notes
         - Land cover file looks like 'GEOSFP.20200101.CN.025x03125.NA.nc'
+        - HEMCO diags file needs to be global, is used to include offshore emissions in state vector
+        - Land cover file and HEMCO diags file need to have the same grid
```
laestrada, line 82: Just looking at the comment -- does this still work with the .5 degree option?

src/utilities/make_state_vector_file.py
```diff
@@ -103,32 +114,34 @@ def make_state_vector_file(
...
     statevector[(statevector.lat < lat_min) | (statevector.lat > lat_max), :] = 0
 
-    # Also set pixels over water to 0
+    # Also set pixels over water to 0, unless there are offshore emissions
     if land_threshold:
-        # Where there is no land, replace with 0
-        land = lc.where(lc > land_threshold)
+        # Where there is neither land nor emissions, replace with 0
+        land = lc.where((lc > land_threshold) | (hd > emis_threshold))
         statevector.values[land.isnull().values] = 0
```
laestrada, line 134: I may be confused what is happening here, but why don't we just set these values to -9999 here rather than a few lines down?

laestrada (approved):
Overall looks good! A few minor clarification questions and a few suggestions.

src/utilities/make_state_vector_file.py
djvaron: Yes it should still work -- will clarify the comment.

config.yml
djvaron: Thanks for the reminders! Done.

src/utilities/make_state_vector_file.py
djvaron: Good point, I made this change and tested on Cannon. Seems to work as expected

djvaron:
Committed changes addressing Lucas's comments [here](https://github.com/geoschem/integrated_methane_inversion/pull/109/commits/2754a937f188fedce6425a29eeb4360a6c1fa8d6).

This feature is untested on AWS and requires the default global HEMCO diag files to be uploaded to `gcgrid`.

laestrada:
@djvaron is this ready to be merged to dev?

PR #112, opened 2023-05-02
remove space before SBATCH settings

nicholasbalasus:


laestrada (approved):
As discussed, looks good.

PR #117, opened 2023-05-12
Add .readthedocs.yaml file to configure ReadTheDocs builds

yantosca:
## Problem
Recently, a software update on the ReadTheDocs side resulted in a side-effect that caused ReadTheDocs builds to fail.   See https://github.com/readthedocs/readthedocs.org/issues/10290 for details.

## Solution
As a recommendation against this from happening again, ReadTheDocs.org recommends adding a `.readthedocs.yaml` file to the root directory.  This configuration file specifies the paths to the `requirements.txt` and `conf.py` that are used by ReadTheDocs to build the documentation.  It also specifies the versions of python and other tools that are used internally.

Here we add a `.readthedocs.yaml` that is consistent with other repositories in the geoschem organization.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,33 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+#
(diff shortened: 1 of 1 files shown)


laestrada:
Thanks for posting this Bob!

PR #122, opened 2023-05-18
Feature/resource allocation settings

laestrada:
This PR allows users to configure the resources (memory, time, and cpus) used to deploy sbatch jobs by adding 4 variables to the configuration file:

```
SimulationCPUs: 24
SimulationMemory: 16000
JacobianCPUs: 1
JacobianMemory: 4000
RequestedTime: "0-6:00"
```

The simulation resources are applied to sbatch jobs that are not run in parallel. Jacobian resources are applied in the initiation of the jacobian simulations.


src/utilities/common.sh
```diff
@@ -0,0 +1,35 @@
...
+        cp "${InversionPath}/imi_output.log" "${OutputPath}/${RunName}/imi_output.log"
+    fi
+    exit 1
+}
+
+# Description: remove sbatch memory and cpu resources
+# remove time too on AWS
+# Usage:
+#   remove_sbatch_headers runscript.run
+remove_sbatch_headers() {
```
msulprizio, line 26: Is there a reason we need this? Can we simply remove those lines from the relevant run scripts (e.g. [ch4_run.template](https://github.com/geoschem/integrated_methane_inversion/blob/main/src/geoschem_run_scripts/ch4_run.template))?

msulprizio (commented):
This PR looks good, but I'm wondering if instead of calling `remove_sbatch_headers` in several places we could simply remove those lines from the run scripts permanently.

Issue #123, opened 2023-05-19
ncmax command is too resource intensive


laestrada:
These two lines seem to be limiting the system requirements capable of running the IMI:

```
ncmax() { ncap2 -O -C -v -s "foo=${1}.max();print(foo)" ${2} ~/foo.nc | cut -f 3- -d ' ' ; }
nElements=$(ncmax StateVector ${RunDirs}/StateVector.nc)
```
Running this on a c5a.8xlarge (32 vCPUs and 64 GB memory)  EC2 instance fails, while a c5.9xlarge (36 vCPUs and 72GB memory) succeeds. We should update this to use a less resource-intensive function for calculating the number of statevector elements.

laestrada:
@msulprizio My feeling is that we could write a short python script to extract the statevector size, but do you have any other ideas via shell commands?

msulprizio:
@laestrada That may be easiest. I was thinking we could output that value from the python script that generates the state vector, but that doesn't help in the case of a user-provided state vector file.

laestrada:
Fixed with #124

PR #124, opened 2023-05-20
Bugfix/get num statevector elements

laestrada:
This PR replaces the `ncmax` and `ncmin` functions (addressing #123) and puts them in the `common.sh` file. The behavior of the updated function is the same except that it uses python instead of ncap2. ncap2 requires significantly more memory and caused a constraint on the amount of memory that could be used to run the IMI.

Additionally, I remove the `setup_imi.sh` file because this is a vestigial file. Setup is now handled by `src/components/setup_component/setup.sh`.


laestrada:
Tested on AWS and confirmed no differences (except for the removal of some trailing whitespace)

msulprizio (approved):
Thanks @laestrada! This PR looks good to merge.

PR #128, opened 2023-06-07
bugfix to catch errors in TROPOMI variables

sabourbaray:

Name: Sabour Baray
Institution: Environment and Climate Change Canada

### Describe the update

In the recently updated RPRO dataset, there appear to be some edge cases where an orbit file can be opened correctly, but still return a hdf5 error when trying to read a variable later on. This can interrupt the process of reading all TROPOMI files for a given time period.

I've shared another [script](https://github.com/sabourbaray/download_copernicus_py) that validates our local TROPOMI cache with the Copernicus repository by comparing exact filesizes, so I've double checked the files in our download process. It may be the case that these are files that could be problematic for any user.

Moving the error-catch to include all variables fixes this problem for me.

### DIFF
--- a/src/inversion_scripts/operators/TROPOMI_operator.py
+++ b/src/inversion_scripts/operators/TROPOMI_operator.py
@@ -455,87 +455,88 @@ def read_tropomi(filename):
     # Initialize dictionary for TROPOMI data
     dat = {}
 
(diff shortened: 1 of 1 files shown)


laestrada (approved):
Looks good -- thanks for catching this Sabour!

PR #131, opened 2023-06-12
bugfix to only turn on extra observation switches for prior and posterior runs

sabourbaray:

Name: Sabour Baray
Institution: Environment and Climate Change Canada

### Describe the update

For the extra observation operators (GOSAT, TCCON, AIRS, Planeflight), currently the IMI switches them on in the template run, which is used to update the geoschem_config.yml files in all run directories. Users will likely only use these comparisons for the base run and posterior run, so we can save computational cost by not running these operators for each jacobian run.

### DIFF
--- a/src/components/jacobian_component/jacobian.sh
+++ b/src/components/jacobian_component/jacobian.sh
@@ -116,15 +116,15 @@ setup_jacobian() {
 		sed -i "s/$OLD/$NEW/g" geoschem_config.yml
 	fi
 	if "$PLANEFLIGHT"; then
-	mkdir -p Plane_Logs
-	sed -i "/planeflight/{N;s/activate: false/activate: true/}" geoschem_config.yml
+		mkdir -p Plane_Logs
+		sed -i "/planeflight/{N;s/activate: false/activate: true/}" geoschem_config.yml
 	
-	OLD="flight_track_file: Planeflight.dat.YYYYMMDD"
-	NEW="flight_track_file: Planeflights\/Planeflight.dat.YYYYMMDD"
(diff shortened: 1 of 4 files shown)


src/components/jacobian_component/jacobian.sh
```diff
@@ -98,6 +98,36 @@ setup_jacobian() {
...
+		NEW="TCCON: true"
+		sed -i "s/$OLD/$NEW/g" geoschem_config.yml
+    	fi
+    	if "$AIRS"; then
+		OLD="AIR: false"
+		NEW="AIR: true"
+		sed -i "s/$OLD/$NEW/g" geoschem_config.yml
+	fi
+	if "$PLANEFLIGHT"; then
+	mkdir -p Plane_Logs
```
laestrada, line 119: nitpick, but the spacing is inconsistent here

laestrada (commented):
@sabourbaray This is a great idea. Thanks for making the PR. My only comment is that perhaps we should make this a function (eg. `activate_observations`) that lives in either `common.sh` or `setup.sh`. Then call the function in both `jacobian.sh` and `posterior.sh`.

sabourbaray:
> @sabourbaray This is a great idea. Thanks for making the PR. My only comment is that perhaps we should make this a function (eg. `activate_observations`) that lives in either `common.sh` or `setup.sh`. Then call the function in both `jacobian.sh` and `posterior.sh`.

Hi Lucas, I've added both of the suggested changes. The function `activate_observations` is in `setup.sh` where I think it will be easier to find when setting up the directories.

laestrada (approved):
Looks good to merge -- thanks Sabour!

PR #133, opened 2023-06-14
Feature/kalman filter refactor

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
- This is an initial attempt at refactoring the original kalman filter branch to utilize new imi components and eliminate the need for a separate run_kf.sh script

TODOs include:
- verifying the specified period of time splits into neat kalman chunks based on the frequency
-  Additional updates to the documentation
- make an additional notebook to visualize resulting variability (maybe copied into `kf_inversions`)

Testing:
- basic 2-week kf runs with no errors (1-week chunks)
- Normal 2-week inversion (without `KalmanMode: true`) runs with no errors 

@djvaron 
it would be good to compare output between your science branch and this one


LICENSE.md
djvaron, line 1: Was deleting the license intentional?

djvaron, line 1: This deletion was probably unintentional.

config.yml
djvaron, line 1: The `FirstPeriod` variable seems to be missing.

djvaron: Noticed the new `EndDate: 20180601` setting, which with `UpdateFreqDays: 7` means that the default is 4 consecutive weekly inversions. Seems reasonable.

run_imi.sh
```diff
@@ -72,8 +73,11 @@ if "$SafeMode"; then
        ([ -d "${RunDirs}/posterior_run" ] && "$SetupPosteriorRun"); then
         
         printf "\nERROR: Run directories in ${RunDirs}/"
-	printf "\n   already exist. Please change RunName or change the"
-	printf "\n   Setup* options to false in the IMI config file.\n"
+        printf "\n   already exist. Please change RunName or change the"
+        printf "\n   Setup* options to false in the IMI config file.\n"
+        printf "\n  To proceed, and overwrite existing run directories, set"
+        printf "\n  SafeMode in the config file to false.\n" 
+        printf "\nIMI $RunName Aborted\n"
```
djvaron, line 80: Duplicate line `printf "\nIMI $RunName Aborted\n"`

src/components/jacobian_component/jacobian.sh
```diff
@@ -74,8 +74,8 @@ setup_jacobian() {
 	    RestartFile=${RestartFilePrefix}${StartDate}_0000z.nc4
 	    ln -s $RestartFile Restarts/GEOSChem.Restart.${StartDate}_0000z.nc4
 	    if "$UseBCsForRestart"; then
-		sed -i -e "s|SpeciesRst|SpeciesBC|g" HEMCO_Config.rc
-            fi
+		    sed -i -e "s|SpeciesRst|SpeciesBC|g" HEMCO_Config.rc
+        fi
```
djvaron, line 78: Looks like the indent formatting got a little confused here.

src/components/jacobian_component/jacobian.sh
```diff
@@ -122,22 +128,55 @@ setup_jacobian() {
...
 
-    printf "\n=== DONE JACOBIAN SIMULATIONS ===\n"
-    jacobian_end=$(date +%s)
+        if "$KalmanMode"; then
+            precomputedSensiCache=${ReferenceRunDir}/kf_inversions/period${i}/data_sensitivities
+        else
+            # Need this alternative when run_imi.sh and run_kf.sh get combined... TODO
+            precomputedSensiCache=${ReferenceRunDir}/inversion/data_sensitivities
+        fi
+        # mv rather than rm, to prevent accidental deletion of original data_sensitivities/ ?
```
djvaron, line 161: Makes sense. Just need to remember this for eventual clean-up options :)

src/components/kalman_component/kalman.sh
```diff
@@ -0,0 +1,151 @@
...
+    python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
+
+    # Create unit scale factor file
+    python ${InversionPath}/src/components/kalman_component/make_unit_sf.py $StateVectorFile $RunDirs; wait
+
+    # Create directory to archive prior scale factors for each inversion period
+    mkdir -p ${RunDirs}/archive_sf
+
+    # Number of state vector elements
+    nElements=$(ncmax StateVector ${StateVectorFile})
```
djvaron, line 31: You wrote a script to replace `ncmax`, right?

src/components/kalman_component/kalman.sh
```diff
@@ -0,0 +1,151 @@
...
+#   run_kf
+run_kf() {
+
+    if ("$DoJacobian" && "$DoInversion" && "$DoPosterior"); then
+
+        # First run the Preview if necessary to get prior emissions
+        # needed for prepare_sf.py
+        if [[ ! -d ${RunDirs}/preview_run/OutputDir ]]; then
+            printf "\nPreview Dir not detected. Running the IMI Preview as a prerequisite for Kalman Mode.\n"
+            run_preview
```
djvaron, line 45: Can this component actually see `run_preview` ?

src/components/kalman_component/kalman.sh
```diff
@@ -0,0 +1,151 @@
...
+        JacobianRunsDir="${RunDirs}/jacobian_runs"
+        PriorRunDir="${JacobianRunsDir}/${RunName}_0000"
+        PosteriorRunDir="${RunDirs}/posterior_run"
+        InversionDir="${RunDirs}/inversion_template"
+
+        PeriodsFile="${RunDirs}/periods.csv"
+        nPeriods=$(($(wc -l < ${PeriodsFile}) - 1))
+
+        # by default, start with period 1
+        if [[ "x${FirstPeriod}" == "x" ]]; then
```
djvaron, line 57: Is this why there is no `FirstPeriod` setting in the default config file? If you don't include it, the IMI knows to start at period 1, and if you want to start at another period then you add the setting?

src/components/kalman_component/kalman.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    ##=======================================================================
+    ##  Submit all Jacobian simulations OR submit only the Prior simulation
+    ##=======================================================================
+    
+    # run jacobian simulation for the given period
+    run_jacobian
+
+    # run inversion for the given period
+    run_inversion
```
djvaron, line 109: I must be forgetting how the components (including preview, which I commented on above) can see each other from their different directories.

src/components/kalman_component/kalman.sh
```diff
@@ -0,0 +1,151 @@
...
+            xstr="${x}"
+        fi
+        ln -sf ${PosteriorRunDir}/Restarts/GEOSChem.Restart.${EndDate_i}_0000z.nc4 ${JacobianRunsDir}/${RunName}_${xstr}/Restarts/.
+    done
+    echo "Copied posterior restart to $((x-1)) Jacobian run directories for next iteration"
+
+    cd ${InversionPath}
+
+    # Delete unneeded daily restart files from Jacobian and posterior directories
+    # python ${InversionPath}/src/components/kalman_component/cull_restarts.py $JacobianRunsDir $PosteriorRunDir $StartDate_i $EndDate_i
```
djvaron, line 146: This is where we could call the `cull_restarts()` function. I can't remember why I commented it out -- perhaps just in testing my original branch...

config.yml
laestrada: `FirstPeriod` is now an optional variable that defaults to 1 if not present in `config.yml`. Before merging I think we should set the config back to 1-week default with `KalmanMode: false`, since I view the kalman filter as an "advanced option"

src/components/jacobian_component/jacobian.sh
laestrada: Yeahhhhh

src/components/kalman_component/cull_restarts.py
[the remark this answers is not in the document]
laestrada: Perhaps in the future cleanup options?

src/components/kalman_component/kalman.sh
laestrada: Yup, but my replacement shares the same name and formatting as the old one -- so this is correct!

src/components/kalman_component/kalman.sh
laestrada: I believe so. Think of the `source` command as pasting all the functions in the corresponding file that called `source kalman.sh`. Because `run_imi.sh` sources both preview.sh and kalman.sh, they should be able to see each other. I should probably check though

src/components/kalman_component/kalman.sh
laestrada: exactly

src/components/kalman_component/kalman.sh
laestrada: see my comment above

src/components/kalman_component/kalman.sh
laestrada: I'll put this back in!

PR #140, opened 2023-06-23
Bugfix: Fix DOFS calculation

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
We noticed a bug in the visualization notebook DOFS calculation. This only occurs when using clustering. In `make_gridded_posterior.py` we map the calculated sensitivity values to the corresponding state vector elements. Eg. if element 3 has a sensitivity of .5, and is composed of 10 grid cells, then all 10 grid cells are set to .5. However, in `visualization_notebook.ipynb` the way we calculate DOFS is `np.nansum(avkern_ROI.values)` which sums every .25 degree grid cell. Per the example, we are effectively overcounting the sensitivity of element 3 by 10x. 

To fix this we simply use the un-gridded inversion result from `inversion_result.nc` to calculate DOFS.

Credit to Alex Oort Alonso for flagging this bug!

Confirmed that this fix works in the Permian test case (with no clustering).

### DIFF
--- a/src/notebooks/visualization_notebook.ipynb
+++ b/src/notebooks/visualization_notebook.ipynb
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


nicholasbalasus (approved):
Looks good to me. Thanks, Lucas!

PR #142, opened 2023-06-28
Feature - create IMI docker image

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This update:
- Establishes the build scripts and documentation for using the IMI docker image
- Adds functionality to automatically zip and upload output files to s3 as a .tar.gz archive
- automatically run visualization notebook and export them to html file for ease of access
- move `DOFSThreshold` check to within `imi_preview.py` to eliminate `bc` dependency


.github/workflows/ecr_deploy.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy to ECR
+
+on:
+  push:
+    branches: [ feature/imi_dockerfile ]
```
laestrada, line 5: change to dev

PR #143, opened 2023-06-29
Bugfix/enable nc4 compression

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Update all `to_netcdf()` calls to use compression.
I set `complevel` to 9, which is the highest setting available. Perhaps this should just be a function.

### DIFF
--- a/envs/Harvard-Cannon/config.harvard-cannon.yml
+++ b/envs/Harvard-Cannon/config.harvard-cannon.yml
@@ -146,14 +146,14 @@ RestartDownload: false
 
 ## Path to initial GEOS-Chem restart file + prefix
 ##   ("YYYYMMDD_0000z.nc4" will be appended)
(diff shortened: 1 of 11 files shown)


nicholasbalasus (approved):
Hi Lucas - I ran this on Cannon (one `dev` and one `bugfix/enable-nc4-compression`) to see what the differences in run time and directory size would be. Everything was on huce_cascade and both run_imi.sh scripts were given 16 GB. Results are promising, but who knows if these will scale from our test case. Anyway, looks good!

`dev` (no compression)
~37 minutes
12 GB

`bugfix/enable-nc4-compression` (yes compression)
~38 minutes
4.1 GB

laestrada:
@nicholasbalasus wow, that is an impressive difference!

PR #145, opened 2023-07-07
Add option to specify region by name

alexoort:

Name: Alex Oort Alonso
Institution: Harvard 

### Describe the update

This update enables users to select their inversion area by typing a country's name. This makes running inversions over entire countries (for national emission tracking) far easier. Previous versions required a user to download a shapefile from an online database and then convert this shapefile into a statevector file through a jupyter notebook.


alexoort:
Also includes changes from the previous pull request, which relate to incorporating plume data into the clustering algorithm

src/utilities/make_state_vector_file.py
```diff
@@ -181,26 +190,321 @@ def make_state_vector_file(
     # Save
     if save_pth is not None:
         print("Saving file {}".format(save_pth))
-        ds_statevector.to_netcdf(
-            save_pth,
-            encoding={
-                v: {"zlib": True, "complevel": 9} for v in ds_statevector.data_vars
-            },
-        )
+        ds_statevector.to_netcdf(save_pth)
```
laestrada, line 193: You are reverting the compression here

src/utilities/make_state_vector_file.py
```diff
@@ -181,26 +190,321 @@ def make_state_vector_file(
...
+
+        found = False  # Flag to track if the ROI is found in the database
+
+        # Iterate over each 'option' tag in the parsed HTML content
+        for country in parser.find_all('option'):
+            if country.string == ROI:  # Check if the country name matches the given ROI
+                countryInfo = str(country)
+                # Extract the country code from the HTML string
+                countryCode = countryInfo.split("=\"")[1].split("_")[0]
+                print(countryCode)  # Print the country code
```
laestrada, line 226: Either remove the print statement or make it more informative. `print(f"{ROI} country found in database with country code: {countryCode}")`

src/utilities/make_state_vector_file.py
```diff
@@ -181,26 +190,321 @@ def make_state_vector_file(
...
+        countryCode    [str]  : Three-letter country code (e.g. USA) of the ROI
+
+    Returns
+        shp_name        [str] : Path to the dowloaded shapefile
+    """
+    shp_url = f"https://geodata.ucdavis.edu/gadm/gadm4.1/shp/gadm41_{countryCode}_shp.zip"  # URL for country shapefile
+
+    try:
+        # Send a GET request to the shapefile URL with a timeout of 5 seconds
+        rshp = requests.get(shp_url, allow_redirects=True, timeout=5)
```
laestrada, line 257: You said you were getting timeout errors with the downloads? Some of the files are reasonably large, maybe increase the timeout to 30 seconds?

laestrada (changes requested):
This looks great overall and is very convenient/useful, since I think generating shapefiles can be a pain point for many people. Also, nice commenting and style!

The main issue is that I think it is confusing to have createAutomaticRectilinearStateVector be true and ignore it in favor of the country. I think you should update [this line](https://github.com/geoschem/integrated_methane_inversion/blob/f0744675a1914c23d8f4d59836d0522e2d59c554/src/components/setup_component/setup.sh#L144) to check that (createAutomaticRectilinearStateVector is false and RegionOfInterest  is true) or (createAutomaticRectilinearStateVector is true and RegionOfInterest is false/unspecified). You can even move that logic into this python script if that is easier.

src/utilities/make_state_vector_file.py
```diff
@@ -485,21 +500,23 @@ def make_country_state_vector_file(config_path, land_cover_pth, hemco_diag_pth,
...
+        elif (config["RegionOfInterest"] is None) and config["CreateAutomaticRectilinearStateVectorFile"] == True:
+            print("Creating Rectilinear StateVector File")
             make_rectilinear_state_vector_file(
             config_path, 
             land_cover_pth, 
             hemco_diag_pth, 
             save_pth,
             )
+        else:
+            print("ERROR: The method of creating a statevector was not specified correctly.")
```
laestrada, line 519: I think you should [raise](https://www.w3schools.com/python/ref_keyword_raise.asp) an error here rather than just a print statement.

src/utilities/make_state_vector_file.py
```diff
@@ -485,21 +500,23 @@ def make_country_state_vector_file(config_path, land_cover_pth, hemco_diag_pth,
...
     except KeyError:
```
laestrada, line 520: Im a bit confused why this is in a try except? What is the exact exception case? It seems like we would want it to fail rather than `make_rectilinear_state_vector_file` If an exception occurs. Usually I try to only use try and except in handle very specific exceptions where I can anticipate the cause of a failure case eg. failing to read a file because it is corrupted.

laestrada (commented):
More minor comments! I am curious how well ruamel imitates the original config file? Does git notice any changes aside from the expected changes to the shapefile path?

alexoort:
Pretty sure that ruamel exactly copies the file's contents, so shouldn't be detected by git. Also have made suggested changes!

Issue #155, opened 2023-08-01
Hardcoding vertical levels in write_boundary.py


eastjames:
In `src/write_BCs/write_boundary.py`, a previously calculated bias correction is applied to GEOS-Chem BoundaryCondition files. The same correction in units of ppb is applied to all model levels. However, in `src/write_BCs/write_boundary.py`, the number of vertical model levels is hardcoded to be 47. I know this is the IMI standard configuration, but the option is given for a user to write boundary conditions from their own GEOS-Chem runs which could have a different configuration than standard IMI, which opens up the possibility that the bias correction would not be applied to all the model levels if a 47 level configuration wasn't used.

Here is the original code:
https://github.com/geoschem/integrated_methane_inversion/blob/b49ce2ffe5cd72ff6ce094b26c62671c88ddc8fe/src/write_BCs/write_boundary.py#L47C1-L49C77

I think a simple fix would be to changes lines 47-49 from:
```python
for t in range(orig_data.shape[0]):
    for lev in range(47):
        orig_data[t, lev, :, :] = orig_data[t, lev, :, :] - Bias
```

to 
```python
orig_data = orig_data - Bias
```

nicholasbalasus:
Closed by #156

Issue #148, opened 2023-07-13
IMI BCs not reproducible with current replace_outliers scheme


laestrada:
### Name and Institution (Required)

Name: Lucas Estrada
Institution: Harvard ACMG

### Description of your issue or question
The current iteration of calculate_bias.py  (`v2023-06`) uses a threshold value that removes bias corrections greater than the threshold to prevent anomalously high bias corrections. However, this threshold is determined by finding the 99th percentile of bias. The percentile changes depending on the input start and end dates, which inhibits reproducibility for the Boundary conditions.

For the next iteration of BCs we should either remove the outlier replacement or update it to use a set amount of time eg. 1-month. Thanks @nicholasbalasus  for finding this bug.

nicholasbalasus:
When we fix this, we also make the simple change to only do this when generating for April 2018:
https://github.com/geoschem/integrated_methane_inversion/blob/5ac2ac6b814e1e78771933ec2da6cefa99662b6b/src/write_BCs/run_BCs.sh#L78-L79

nicholasbalasus:
Another thing that is relevant when not generating for the full period (April 2018 - present): if we are going to keep our temporal averaging as is, we need to run GC and get TROPOMI data for ~20 days before the start date requested. Right now, if you do:


startdate: 20230401
enddate: 20230530

It will only run GC and get TROPOMI for 2023.04.01-2023.06.20. It should get it for 2023.03.11-2023.06.20 (or so).

nicholasbalasus:
Closed by #156

Issue #147, opened 2023-07-13
Enforce (and download) GCClassic version for the IMI


laestrada:
### Name and Institution (Required)

Name: Lucas Estrada
Institution: Harvard ACMG

### New IMI feature or discussion
WHen downloading the IMI for local cluster use, the user separately need to git clone GCClassic using the correct version that the IMI expects. We should update the IMI to either have GCClassic as a submodule in the repository (using the correct version) or have the IMI download and install GCClassic at runtime, similar to how @nicholasbalasus set up the BC scripts. If the latter, it should check if GCClassic is already installed, and, if so, skip the download.

msulprizio:
This is now done in #156. I will close out this issue.

Issue #146, opened 2023-07-13
Update IMI Boundary Conditions to use 2x2.5 degree simulation


laestrada:
### New IMI feature or discussion

Name: Lucas Estrada
Institution: Harvard ACMG

### Description of your issue or question
Our boundary condition scripts currently use a 4x5 degree resolution simulation, but work by Todd and others suggests that using 2x2.5 degree simulations vastly improves transport and vertical distribution in the troposphere and stratosphere when compared to observations.

We should update the BCs to use the 2x2.5 degree simulation when we update to version 2.0/

nicholasbalasus:
Closed by #156

PR #156, opened 2023-08-01
[Draft] low latency boundary condition generation

nicholasbalasus:

Name: Nick Balasus

### Describe the update

This draft PR will include updates to how we generate boundary conditions (likely `v2023-08` that will accompany the IMI 2.0). This includes some bug fixes but mostly changes that will allow us to generate boundary conditions with lower latency.

Changes that will impact the output:
- [x] switch to 2° x 2.5° (#146)
- [x] upgrade to GEOS-Chem v14.2.1 and ensure consistency with IMI (#147)
- [x] sample GEOS-Chem SpeciesConcVV_CH4 and LevelEdgeDiagn instantaneously instead of time-averaged (previous bug)
- [x] remove the dropping of outliers (#148)
- [x] use a restart file for 1 April 2018 from Todd's simulation (from NOAA surface data)
- [x] reduce the temporal smoothing of biases to -15 days (instead of +/- 15 days)
- [x] drop observations over water and south of 60°S
- [x] explore data density minimums (when regridding and when defining a zonal mean bias)
- [x] checking that our TROPOMI operator is behaving as expected (#110)

Changes that will not impact the output:
- [x] modify the default GEOS-Chem config files with `sed` instead of providing config files
- [x] add flexibility for number of vertical layers (#155)


nicholasbalasus:
I am only requesting the review from @laestrada, but I will tag others in places they might want to chime in (thanks @djvaron, @msulprizio, @toddmooring, @eastjames, @megan-he!). This should close the five issues linked above.

This update uses GCClassic 14.2.1 to generate boundary conditions at 2.0° x 2.5°. They will be more appropriately named `v2023-09` (my branch name with `v2023-08` was too ambitious it seems). I will generate the full record (and compare to the 4.0° x 5.0° `v2023-06` boundary conditions) once this code looks good and 14.2.1 is officially released.

I have tested two things here:

**(1) The default IMI test case on Cannon.**

I tested this because GCClassic is now cloned as part of the setup for IMI (getting rid of the need for users to clone it themselves). This is the bash script used to test: [test-default-imi.txt](https://github.com/geoschem/integrated_methane_inversion/files/12561135/test-default-imi.txt). 

It works - results here if you're interested: [visualization_notebook.html.zip](https://github.com/geoschem/integrated_methane_inversion/files/12561477/visualization_notebook.html.zip)

**(2) The reproducibility of our boundary conditions generation.**

I generated three sets of boundary conditions: test 1 of 20180401-20180630, test 2 of 20180515-20180531 (using a restart generated by test 1), test 3 of 20180615-20180630 (using a restart generated by test 2). This is the bash script used to test: [test-bcs.txt](https://github.com/geoschem/integrated_methane_inversion/files/12561158/test-bcs.txt). The boundary conditions from test 2 and test 3 are consistent with those generated in test 1, as is tested with this simple Python code:

```
import xarray as xr
import numpy as np
from datetime import date, timedelta

basedir = "/n/holyscratch01/jacob_lab/nbalasus/"

test2_dates_to_check = [(date(2018, 5, 15) + timedelta(days=i)).strftime("%Y%m%d") for i in range(17)]
test3_dates_to_check = [(date(2018, 6, 15) + timedelta(days=i)).strftime("%Y%m%d") for i in range(16)]

for date in test2_dates_to_check:
    with xr.open_dataset(basedir + f"imi-test1/blended-boundary-conditions/GEOSChem.BoundaryConditions.{date}_0000z.nc4") as ds1,\
         xr.open_dataset(basedir + f"imi-test2/blended-boundary-conditions/GEOSChem.BoundaryConditions.{date}_0000z.nc4") as ds2:
        abs_diff = 1e9*np.abs((ds1["SpeciesBC_CH4"] - ds2["SpeciesBC_CH4"]).values)
        print(f"{date} -> {np.max(abs_diff)}")
        assert np.max(abs_diff) < 0.01, f"Diff is greater than 0.01 ppb for {date}"

for date in test3_dates_to_check:
    with xr.open_dataset(basedir + f"imi-test1/blended-boundary-conditions/GEOSChem.BoundaryConditions.{date}_0000z.nc4") as ds1,\
         xr.open_dataset(basedir + f"imi-test3/blended-boundary-conditions/GEOSChem.BoundaryConditions.{date}_0000z.nc4") as ds2:
        abs_diff = 1e9*np.abs((ds1["SpeciesBC_CH4"] - ds2["SpeciesBC_CH4"]).values)
        print(f"{date} -> {np.max(abs_diff)}")
        assert np.max(abs_diff) < 0.01, f"Diff is greater than 0.01 ppb for {date}"
 ```

File: src/inversion_scripts/operators/TROPOMI_operator.py [hunk 1]
```diff
@@ -323,6 +323,11 @@ def apply_tropomi_operator(
                 coords[1] + dlat / 2,
                 coords[1] + dlat / 2,
             ]
+            # If this is a global 2.0 x 2.5 grid, extend the eastern-most grid cells to 180 degrees
+            if (dlon == 2.5) & (coords[0] == 177.5):
+                for i in [1,2]:
+                    geoschem_corners_lon[i] += dlon / 2
+
```
nicholasbalasus, line 330: @megan-he @eastjames this is not a big deal, just flagging it because it's kind of weird. To ensure assert abs(sum(overlap_area)-polygon_tropomi.area)/polygon_tropomi.area < 0.01 is true (that is, that we find a GEOS-Chem grid cell for all TROPOMI pixels which should be true on a global grid), I had to expand the eastern-most grid boxes on the 2.0 x 2.5 grid to 180 degrees (because the centers go from -180 to 177.5, so 178.75-180 did not exist!) Might be relevant elsewhere, but not a big deal.

src/inversion_scripts/utils.py
```diff
@@ -293,6 +294,8 @@ def filter_tropomi(tropomi_data, xlim, ylim, startdate, enddate):
         & (tropomi_data["time"] <= enddate)
         & (tropomi_data["qa_value"] >= 0.5)
         & (tropomi_data["longitude_bounds"].ptp(axis=2) < 100)
+        & (tropomi_data["surface_classification"] != 1)
+        & (tropomi_data["latitude"] > -60)
```
nicholasbalasus, line 298: @djvaron back in that long email thread about new boundary conditions, we decided to drop data over the water and south of 60S. This means it will be dropped in inversions as well (to be consistent with boundary conditions). Is that okay?

src/write_BCs/GEOSChem.Restart.20180401_0000z.nc4
nicholasbalasus, line 1: @toddmooring I am using the file from `/n/jacob_lab/Users/tmooring/GCClassic-correcttropopause_archdirs/gc_2x25_merra2_CH4_ct_CLSBCr_prod1/Restarts/GEOSChem.Restart.20180401_0000z.nc4` - can you confirm that is correct?

src/inversion_scripts/operators/TROPOMI_operator.py
megan-he: Thanks Nick! I had to change the max longitude of my state vector from 180 to 177.5 previously to avoid index mismatching, so I think this should resolve that issue.

src/components/setup_component/setup.sh
```diff
@@ -127,7 +127,15 @@ setup_imi() {
...
 
+    # Clone version 14.2.1 of GCClassic
     # Define path to GEOS-Chem run directory files
+    cd "${InversionPath}"
+    rm -rf GCClassic
+    git clone https://github.com/geoschem/GCClassic.git
+    cd GCClassic
+    git checkout dac5a54 # most recent dev/14.2.1 @ 1 Sep 2023 12:44 PM (update this once 14.2.1 officially released)
+    git submodule update --init --recursive
+    cd ..
```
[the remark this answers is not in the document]
laestrada: Thanks Nick, I would prefer if there was a check to see if the GCClassic directory is present, rather than deleting and redownloading every time. This allows users more flexibility to modify GEOS-Chem. I think wrapping the above in `if [ ! -d "GCClassic" ]; then` would be sufficient.

src/write_BCs/write_boundary_conditions.py
```diff
@@ -0,0 +1,236 @@
...
+    nan_value_filler_3d = bias.copy() * np.nan
+    for i in range(len(LON)):
+        nan_value_filler_3d[:,:,i] = nan_value_filler_2d
+
+    # Use these values to fill NaNs
+    bias = bias.fillna(nan_value_filler_3d)
+
+    print(f"Average bias (GC-TROPOMI): {bias.mean().values:.2f} ppb\n")
+
+    # If there are still NaNs (this will happen when TROPOMI data is missing), use 0.0 ppb as the bias but warn the user
```
laestrada, line 195: Nice, this is a good idea

src/write_BCs/write_boundary_conditions.py
```diff
@@ -0,0 +1,236 @@
...
+    - average the TROPOMI XCH4 and GEOS-Chem XCH4 to a (2.0 x 2.5) grid for each day
+(2) Make a gridded (2.0 x 2.5 x daily) field of the bias between TROPOMI and GEOS-Chem
+    - subtract the TROPOMI and GEOS-Chem grids from part 1 to get a starting point for the bias
+    - smooth this field spatially (5 lon grid boxes, 5 lat grid boxes) then temporally (15 days backwards)
+    - fill NaN values with the latitudinal average at that time
+        - for a latitudinal average to be defined, there must be >= 30 grid cells at that latitude
+        - when a latitudinal average cannot be found, the closest latitudinal average is used
+(3) Write the boundary conditions
+    - using the bias from Part 2, subtract the (GC-TROPOMI) bias from the GC boundary conditions
+"""
```
laestrada, line 46: The content of this looks good. I do wonder if the main script is getting to be too bloated.. perhaps we should separate steps 1,2, and 3 into functions?

laestrada (changes requested):
Overall looks good, some minor changes requested.

nicholasbalasus:
Thanks, @laestrada! Everything is addressed now (except for `isRegional`, which will need to wait for now). Testing now.

src/inversion_scripts/operators/TROPOMI_operator.py
eastjames: Looks good @nicholasbalasus!

src/write_BCs/README.md
```diff
@@ -11,6 +11,7 @@
    - `Partition`       - which partition to run the jobs on.
    - `restartFilePath` - restart file for GEOS-Chem.
       - if your simulation starts on 1 April 2018, this won't be used (`GEOSChem.Restart.20180401_0000z.nc4` will).
+         - this file comes from a simulation from a CH4 simulation by Todd Mooring that is constrained by NOAA surface observations.
```
laestrada, line 14: maybe remove the first from a simulation ;)

src/write_BCs/write_boundary_conditions.py
```diff
@@ -233,4 +208,38 @@ def apply_tropomi_operator_to_one_tropomi_file(filename):
...
+        - fill NaN values with the latitudinal average at that time
+            - for a latitudinal average to be defined, there must be >= 30 grid cells at that latitude
+            - when a latitudinal average cannot be found, the closest latitudinal average is used
+    (3) Write the boundary conditions
+        - using the bias from Part 2, subtract the (GC-TROPOMI) bias from the GC boundary conditions
+    """
+
+    daily_means = part1(satelliteDir, start_time_of_interest, end_time_of_interest)
+    bias = part2(daily_means)
+    part3(bias)
```
laestrada, line 245: Can you make the function names descriptive? Something like: `create_daily_means()`, `calculate_bias()`, and `write_bias_corrected_files()`

nicholasbalasus:
Thanks @laestrada! Let me know if this is ready to squash + merge.

laestrada (approved):
Thanks Nick, this looks good to me!

src/components/setup_component/setup.sh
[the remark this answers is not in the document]
msulprizio: @nicholasbalasus I'm not sure a warning would be good enough here because it might get overlooked. In order to ensure the correct version of GCClassic is used, we should check for the version number in `GCClassic/CMakeLists.txt`. In that file, you should see the text `VERSION 14.2.1`. If it doesn't match the expected version then the setup script can either stop with an error (along the lines of the message you have above) or preferably print a message telling the user GCClassic is being updated to X.Y.Z and automatically update the GCClassic repo and submodules.

Also, rather than checking out a specific commit, I would recommend using the version tag (e.g. `git checkout 14.1.1`). We don't recommend pegging it to a development version in general, but if you must then you should use the branch name instead of the commit ID (e.g. `git checkout dev/14.2.1`). We can change it to use the tag once the version is officially released/tagged.

toddmooring:
Hi everyone but especially @nicholasbalasus and @laestrada . Apologies for the delayed response. I just want to make sure I understand how the boundary condition generation is supposed to work:

1) The user picks a start date for the inversion on or after 1 April 2018.
2) If the date is after 1 April 2018, global GEOS-Chem is run forward at 2x2.5/47L to the start date. The initial condition is a 1 April 2018 restart supplied by me, emissions are assumed based on some inventory.
3) From the start date on to the end date, the global 2x2.5/47L run writes boundary condition files.
4) The inversion is performed, using among other things the boundary conditions generated by the above process.

Assuming this is how it works, I have two questions/suggestions:

1) What happened to Daniel's suggestion of only doing a 1-month emissions-driven spinup after initializing with one of my restarts (see his 7/27/23 1:04 p.m. email)? Are we just now relying on the bias correction process (in which we make the GEOS-Chem state look more like TROPOMI) to mitigate biases created by emissions-driven spinups of more than 1 month?

2) Note that I actually now have monthly restarts as late as 1 May 2018 from a **2x2.5/47L** simulation. This simulation was initialized (from a 2x2.5/72L restart) at 1 May 2012, with that restart coming from a 2x2.5/72L simulation started on 1 January 1985. I think you should switch to using a native 47L restart, for two reasons: a) it skips the process of interpolating from 72L to 47L and b) the stratospheric methane climate has had ~6 years to adjust to the 47L structure. Probably not hugely different from the 72L structure, but why not get rid of this source of a potential fake transient if you can?

toddmooring:
Just to be clear, I realize my description of how the boundary condition generation process works glosses over the application of bias corrections to GEOS-Chem fields to make them better match TROPOMI.

nicholasbalasus:
Hi Todd - thanks for looking through this! 

The upshot of the code is GEOS-Chem (2x2.5/47L) is initialized with your 1 April 2018 restart file and run until present. The archived boundary conditions are corrected to look like TROPOMI (starting when the TROPOMI beings on 30 April 2018), which is what we need for nested-grid simulations for regional inversions. 

I think I understand how what you are referring to in (1) would work for making a restart file for a global simulation (that is, run 1 December 2018 with emissions starting with your restart and then correct the 1 January 2019 output with TROPOMI). However, I don't see how this would be practical for generating multiple years of boundary conditions that we need for nested-grid simulations.

For (2), I will switch to the 47 L file if you point me towards it. Thanks!

toddmooring:
I have conveyed the 2x2.5/47L restart to @nicholasbalasus via Slack

laestrada:
@toddmooring Thanks for providing the restart file! Just to further explain Nick's explanation -- we do not expect users to generate the boundary conditions themselves. Instead, we create an archive (available from the washu data portal and aws), which the IMI downloads for the relevant user dates.

Issue #67, opened 2022-09-19
[FEATURE REQUEST] Include capability to optimize annual OH


msulprizio:
For global inversions, we should also include optimization of OH in addition. This will require updates to the GEOS-Chem simulations to perturb OH. These perturbations may be applied globally or separately to the North and South Hemispheres, but for most applications the global perturbation is sufficient.

msulprizio:
This has been addressed by https://github.com/geoschem/integrated_methane_inversion/pull/157

Issue #59, opened 2022-09-19
[FEATURE REQUEST] Add capability for global simulations in the IMI


msulprizio:
Theoretically it is possible to run the IMI on a global grid, as opposed to the default nested-grid domains. However, more testing is needed.

msulprizio:
This was addressed by https://github.com/geoschem/integrated_methane_inversion/pull/157

PR #157, opened 2023-08-04
Add global inversion capability + OH optimization

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard / GCST

### Describe the update

Here we update the IMI to include the capability for global inversions in addition to the preexisting regional capability. This includes adding the option to run with 4x5 and 2x2.5 horizontal resolutions. In addition, we now modify the terminology to use "regional" instead of "nested" for clarity.

**Update 5 Feb 2024:**

This pull request now also includes the OH optimization udpates originally in https://github.com/geoschem/integrated_methane_inversion/pull/173. From that pull request:

>Additional options to optimize for OH have been added to config.yml. This includes the switch for turning this option on (OptimizeOH), PriorErrorOH, and PerturbValueOH. When enabling this option, an additional run will be added to the jacobian runs. Code has also been added to the inversion scripts for including OH optimization.

Closes:

- https://github.com/geoschem/integrated_methane_inversion/issues/59
- https://github.com/geoschem/integrated_methane_inversion/issues/67


src/components/jacobian_component/jacobian.sh
```diff
@@ -29,6 +29,11 @@ setup_jacobian() {
     sed -i -e "s:{START}:0:g" \
            -e "s:{END}:${nElements}:g" \
            -e "s:{InversionPath}:${InversionPath}:g" jacobian_runs/submit_jacobian_simulations_array.sh
+    if [ $MaxSimultaneousRuns -gt 0 ]; then
+	sed -i -e "s:{JOBS}:%${MaxSimultaneousRuns}:g" jacobian_runs/submit_jacobian_simulations_array.sh
+    else
+	sed -i -e "s:{JOBS}::g" jacobian_runs/submit_jacobian_simulations_array.sh
+    fi
```
laestrada, line 36: How does this handle the case where the number of perturbation sims is under the number of MaxSimultaneousRuns? Eg. 71 perturbation sims, but MaxSimultaneousRuns is set to 100? Would this cause a bug/crash of submit_jacobian_simulations_array.sh?

src/components/preview_component/preview.sh
```diff
@@ -9,8 +9,14 @@
 #   run_preview
 run_preview() {
     if ! "$isAWS"; then
-	# Load environment with modules for running GEOS-Chem Classic
-        source ${GEOSChemEnv}
+	if [ ! -f "${InversionPath}/${GEOSChemEnv}" ]; then
```
laestrada, line 12: I wonder if this check should be in run_imi.sh since we would want this behavior not just for the preview, but anywhere.

envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
```diff
@@ -0,0 +1,198 @@
...
+## Path to TROPOMI Data
+DataPathTROPOMI: "/n/holylfs05/LABS/jacob_lab/imi/ch4/tropomi"
+
+## Conda environment file
+## See envs/README to create the Conda environment specified below
+CondaEnv: "imi_env"
+
+## GEOS-Chem environment file (with fortran compiler, netcdf libraries, etc.)
+##   NOTE: Copy your own file in the envs/ directory within the IMI
+GEOSChemEnv: "envs/Harvard-Cannon/gcclassic.rocky+gnu12.minimal.env"
```
laestrada, line 176: Nice, I like having this relative to the `integrated_methane_inversion` dir

src/components/spinup_component/spinup.sh
```diff
@@ -89,8 +89,14 @@ run_spinup() {
     cd ${RunDirs}/spinup_run
 
     if ! "$isAWS"; then
-        # Load environment with modules for compiling GEOS-Chem Classic
-        source ${GEOSChemEnv}
+	if [ ! -f "${InversionPath}/${GEOSChemEnv}" ]; then
```
laestrada, line 92: see comment above

src/components/statevector_component/aggregation.py
```diff
@@ -238,7 +242,7 @@ def get_max_aggregation_level(config, sensitivities, desired_element_num):
         # if there are too few clusters then we set the max aggregation level
         # to either total_native_elements/8 or total_native_elements
         denominator = 8 if desired_element_num > 8 else 1
-        max_aggregation_level = np.ceil(len(sensitivities) / denominator)
+        max_aggregation_level = np.ceil(len(sensitivities) / denominator).astype(int)
```
laestrada, line 245: Good catch here

src/components/statevector_component/statevector.sh
```diff
@@ -11,15 +11,19 @@ create_statevector() {
...
+    else
+	LandCoverFile="${DataPath}/GEOS_${gridDir}/${metDir}/${constYr}/01/${Met}.${constYr}0101.CN.${gridFile}.${LandCoverFileExtension}"
+    fi
+    HemcoDiagFile="${DataPath}/HEMCO/CH4/v2023-04/HEMCO_SA_Output/HEMCO_sa_diagnostics.${gridFile}.20190101.nc"
+	
     if "$isAWS"; then
-	# Download land cover and Hemco diagnostics files
-	s3_lc_path="s3://gcgrid/GEOS_${gridDir}/${metDir}/${constYr}/01/${metUC}.${constYr}0101.CN.${gridRes}.${NestedRegion}.${LandCoverFileExtension}"
+	# Download land cover and HEMCO diagnostics files
+	s3_lc_path="s3://gcgrid/GEOS_${gridDir}/${metDir}/${constYr}/01/${Met}.${constYr}0101.CN.${gridFile}.${RegionID}.${LandCoverFileExtension}"
```
laestrada, line 23: indentation is weird here but maybe just an artifact of github rendering

src/components/template_component/template.sh
```diff
@@ -119,16 +134,22 @@ setup_template() {
...
-    fi
-
     # Remove sample restart file
     rm -f Restarts/GEOSChem.Restart.20190101_0000z.nc4
 
     # Copy template run script
     cp ${InversionPath}/src/geoschem_run_scripts/ch4_run.template .
+
+    if ! "$isAWS"; then
+	if [ ! -f "${InversionPath}/${GEOSChemEnv}" ]; then
```
laestrada, line 144: again, maybe just one check in run_imi.sh

src/inversion_scripts/calc_sensi.py
```diff
@@ -87,6 +87,21 @@ def calc_sensi(
         nlon = len(base_data["lon"])  # 52
         nlat = len(base_data["lat"])  # 61
         nlev = len(base_data["lev"])  # 47
+        base_var = base_data["SpeciesConcVV_CH4"] # Read base data before loop
+
+        # Save this data into numpy array so we don't need to read files in loop
+        pert_datas = []
+         # For each state vector element
+        for e in elements:
```
laestrada, line 95: is the reason for moving this out of process to avoid memory issues or something else?

laestrada (changes requested):
A few minor comments, questions, and requested changes -- nice work!

run_imi.sh
```diff
@@ -197,7 +198,8 @@ cd $InversionPath
 cp $ConfigFile "${RunDirs}/config_${RunName}.yml"
 
 # Upload output to S3 if specified
-cd $InversionPath
-python src/utilities/s3_upload.py $ConfigFile
+if "$S3Upload"; then
+    python src/utilities/s3_upload.py $ConfigFile
+fi
```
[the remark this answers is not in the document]
msulprizio: OK. I can remove it. I originally added it because I was using an older IMI python environment that didn't have boto3 and that caused error messages when calling s3_upload.py (which I didn't need when running from a local cluster anyway). I resolved the error messages by later updating my conda environment.

config.yml
```diff
@@ -158,7 +164,7 @@ OutputPath: "/home/ubuntu/imi_output_dir"
 ## Path to GEOS-Chem input data
 DataPath: "/home/ubuntu/ExtData"
 
-## Environment files
+## Conda anvironment file
```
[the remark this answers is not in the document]
msulprizio: Now fixed in  https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7

config.yml
```diff
@@ -77,9 +76,12 @@ Gamma: 1.0
...
 
 ## Grid
-##   Select "0.25x0.3125" and "geosfp", or "0.5x0.625" and "merra2"
+##   Options are 0.25x0.3125 (GEOSFP only), 0.5x0.625, 2.0x2.5, or 4.0x5.0
 Res: "0.25x0.3125"
-Met: "geosfp"
+
+## Meteorology
+##   Options are GEOSFP or MERRA2
+Met: "GEOSFP"
```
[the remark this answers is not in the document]
msulprizio: I added flexibility to allow for any upper or lower case in commit https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7.

run_imi.sh
```diff
@@ -3,7 +3,8 @@
 #SBATCH -N 1
 #SBATCH -n 1
 #SBATCH -o "imi_output.log"
-
+#SBATCH -t 0-16:00
+#SBATCH --mem=20000
```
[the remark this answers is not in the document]
msulprizio: Now removed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7

run_imi.sh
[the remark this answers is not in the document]
msulprizio: See https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7

src/components/jacobian_component/jacobian.sh
msulprizio: I'm not sure how slurm would handle this. In commit https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7 I put in an error check to ensure MaxSimultaneousRuns is less than the total number of runs (aka nElements).

src/components/preview_component/preview.sh
msulprizio: These lines have been moved to run_imi.sh in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7.

src/components/spinup_component/spinup.sh
msulprizio: These lines have been moved to run_imi.sh in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7.

src/components/statevector_component/statevector.sh
msulprizio: The indentation issues were caused by the use of tabs instead of spaces. I This has been fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7.

src/components/template_component/template.sh
msulprizio: These lines have been moved to run_imi.sh in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/9629c549521385b0709e1aa33c1c158082f6ddc7.

src/inversion_scripts/calc_sensi.py
msulprizio: Meghan He added these lines in the commit titled "bug fixes for 2x2.5, parallelize jacobian and faster calc_sensi" so it would seem these lines were added to speedup that code.

src/components/statevector_component/aggregation.py
msulprizio: Credit goes to Megan He :)

src/utilities/make_state_vector_file.py
```diff
@@ -138,22 +146,24 @@ def make_state_vector_file(
...
+        irows_good = irows[buffer_area > 0]
+        icols_good = icols[buffer_area > 0]
+        coords = [[icols_good[j], irows_good[j]] for j in range(len(irows_good))]
+
+        # K-means
+        X = np.array(coords)
+        kmeans = KMeans(n_clusters=k_buffer_clust, random_state=0).fit(X)
 
     # Assign pixels to state vector
+    # -------------------------------------------------------------------------
```
laestrada, line 166: kmeans gets calles just below this on line 174 -- shouldn't it be included in the if statement? Since kmeans only handles the buffer cells which are not needed for global?

msulprizio: Thanks for catching that! It's now fixed in eee2cc2.

laestrada (approved):
Looks good, thanks for addressing comments. I filled in the config file documentation for the missing variables you pointed out.

src/geoschem_run_scripts/submit_jacobian_simulations_array.sh
```diff
@@ -4,7 +4,7 @@ echo "running {END} jacobian simulations" >> {InversionPath}/imi_output.log
 # remove error status file if present
 rm -f .error_status_file.txt
 
-sbatch --array={START}-{END} --mem $JacobianMemory \
+sbatch --array={START}-{END}{JOBS} --mem $JacobianMemory \
```
laestrada, line 7: What does {JOBS} do here?

msulprizio: @laestrada Just seeing this question now. In the configuration file, there is a new option `MaxSimultaneousRuns` that allows users to submit their Jacobian runs in batches rather than submitting them all to the queue at once. This is especially useful if the number of state vector elements (and therefore Jacobian runs) is large. If `MaxSimultaneousRuns` is greater than zero then `{JOBS}` will be replaced with `%` followed by the number of runs specified. If `MaxSimutaneousRuns` is -1 (or really any value less than zero) then `{JOBS}` will be removed from that line and the entire job array will be submitted to the queue at once.

msulprizio:
This branch now merges in the OH optimization updates originally in #173.

src/inversion_scripts/invert.py
```diff
@@ -207,15 +215,24 @@ def do_inversion(
...
+    if prior_err_oh > 0.0:
+        # add prior error for OH as the last element of the diagonal
+        Sa_diag[-1:] = (1/1000)*prior_err_oh**2 # try this
+        OH_idx -= 1
         
+    # If optimizing boundary conditions, adjust for it in the inversion
+    if bc_optimization:
+        # add prior error for BCs as the last 4 elements of the diagonal
+        if prior_err_oh > 0.0:
+            Sa_diag[-5:-2] = prior_err_bc**2
```
laestrada, line 230: I think that this indexing is not quite correct. If my understanding is correct, it should be selecting element (n-4),(n-3),(n-2),(n-1), whereas I think this indexing only selects (n-4),(n-3),(n-2).

See example:
```
>>> x = np.array([0,1,2,3,4,5,6,7,8,9])
>>> x[-5:-2]
array([5, 6, 7])
```

I think the fix is to make it `Sa_diag[-5:-1] = prior_err_bc**2`

src/inversion_scripts/invert.py
```diff
@@ -226,43 +243,25 @@ def do_inversion(
     # space, so we do not need to add 1
     # xhat = 1 + ratio
     xhat = ratio.copy()
-    xhat[:bc_idx] += 1
+    xhat[:OH_idx] += 1
```
laestrada, line 246: I think this would add 1 to every element except for the nth element, but it should be using `bc_idx`. which takes into account the bc elements and the OH element. Maybe a better variable name would be `scale_factor_idx`, which tracks the index of elements using scale factors?

laestrada: An aside, is the OH optimization a scale factor or in concentration space?

src/inversion_scripts/invert.py
```diff
@@ -226,43 +243,25 @@ def do_inversion(
...
+        
     # Posterior error covariance matrix
     S_post = np.linalg.inv(gamma * KTinvSoK + inv_Sa)
 
     # Averaging kernel matrix
     A = np.identity(n_elements) - S_post @ inv_Sa
     
     # Print some statistics
-    print("Min:", xhat[:bc_idx].min(), "Mean:", xhat[:bc_idx].mean(), "Max", xhat[:bc_idx].max())
+    print("Min:", xhat[:OH_idx].min(), "Mean:", xhat[:OH_idx].mean(), "Max", xhat[:OH_idx].max())
```
laestrada, line 260: Same as above, I think this should use `scale_factor_idx`

src/inversion_scripts/invert_calcJ.py
```diff
@@ -0,0 +1,317 @@
...
+import glob
+import numpy as np
+import xarray as xr
+import yaml
+from scipy.interpolate import interp2d
+from netCDF4 import Dataset
+from src.inversion_scripts.utils import load_obj
+
+
+def do_inversion(
```
laestrada, line 12: This file seems to be nearly the same as `invert.py`, but with the end goal of calculating Ja. I dont think we need this file. My lognormal-errors branch updates invert.py to prion out Ja. @megan-he can you confirm whether we need this or not?

src/inversion_scripts/jacobian.py
```diff
@@ -120,7 +121,8 @@ def apply_operator(operator, params):
 
     # Map GEOS-Chem to TROPOMI observation space
     # Also return Jacobian matrix if build_jacobian=True
-    for filename in sat_files:
+    def process(filename):
+    # for filename in sat_files:
```
laestrada, line 125: remove comment here

src/inversion_scripts/make_gridded_posterior.py
```diff
@@ -65,9 +65,9 @@ def make_gridded_posterior(posterior_SF_path, state_vector_path, save_path):
     lon = gridded_SF["lon"].values
     ds = xr.Dataset(
         {
-            "ScaleFactor": (["lat", "lon"], gridded_SF),
-            "S_post": (["lat", "lon"], gridded_S_post),
-            "A": (["lat", "lon"], gridded_A),
+            "ScaleFactor": (["lat", "lon"], gridded_SF.data),
+            "S_post": (["lat", "lon"], gridded_S_post.data),
+            "A": (["lat", "lon"], gridded_A.data),
```
laestrada, line 70: According to the numpy documentation `ndarray.data` is a `Python buffer object pointing to the start of the array’s data`. This is directly accessing memory and from what I have read can be error prone. I think remove the `.data` here

src/inversion_scripts/run_inversion.sh
```diff
@@ -3,6 +3,9 @@
 #SBATCH -N 1
 #SBATCH -o run_inversion_%j.out
 #SBATCH -e run_inversion_%j.err
+#SBATCH -t 0-6:00
+#SBATCH --mem=20000
+#SBATCH -p seas_compute,shared,huce_intel
```
laestrada, line 8: remove these defaults, this will override the specified values in the IMI config file

src/inversion_scripts/run_inversion.sh
```diff
@@ -71,29 +74,29 @@ fi
 
 printf "Calling postproc_diags.py, FSS=$FirstSimSwitch\n"
 if "$FirstSimSwitch"; then
-    if [[ ! -d ${SpinupDir} ]]; then
-    printf "${SpinupDir} does not exist. Please fix SpinupDir or set FirstSimSwitch to False in run_inversion.sh.\n"
-    exit 1
-    fi
-    PrevDir=$SpinupDir
+   if [[ ! -d ${SpinupDir} ]]; then
```
laestrada, line 77: It looks like the number of spaces for tab is inconsistent with above in this file

run_imi.sh
```diff
@@ -187,17 +201,16 @@ if ! "$KalmanMode"; then
     print_stats
 fi 
 
-# copy output log to run directory for storage
+# Move output log to run directory for storage
 if [[ -f ${InversionPath}/imi_output.log ]]; then
-    cp "${InversionPath}/imi_output.log" "${RunDirs}/imi_output.log"
+    mv -v "${InversionPath}/imi_output.log" "${RunDirs}/imi_output.log"
```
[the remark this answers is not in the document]
msulprizio: I switched it to move because I an old imi_output.log file from a previous inversion was being copied into newer IMI runs even though they crashed. I figured we could either (1) remove the imi_output.log file when submitting a new IMI run or (2) simply move the imi_output.log file after a run is complete. If you'd prefer option 1 (or neither) let me know.

src/components/jacobian_component/jacobian.sh
```diff
@@ -70,74 +80,90 @@ setup_jacobian() {
...
-            PerturbBCValues=$(generate_BC_perturb_values $bcThreshold $x $PerturbValueBCs)
-            sed -i -e "s|CH4_boundary_condition_ppb_increase_NSEW:.*|CH4_boundary_condition_ppb_increase_NSEW: ${PerturbBCValues}|g" \
-                -e "s|perturb_CH4_boundary_conditions: false|perturb_CH4_boundary_conditions: true|g" \
-                -e "s|emission_perturbation_factor: ${PerturbValue}|emission_perturbation_factor: 1.0|g" \
-                -e "s|state_vector_element_number: ${xUSE}|state_vector_element_number: 0|g" geoschem_config.yml
-        fi
-    fi 
+	# BC optimization setup
+	if "$OptimizeBCs"; then
+	    if "$OptimizeOH"; then
```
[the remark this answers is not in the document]
msulprizio: Nope. First there is a check for OptimizeBCs and if that's on it will also check if optimizeOH is on, and if so, it will account for OH as the last element. After that if statement, there's another one for OptimizeBCs.

```
	# BC optimization setup
	if "$OptimizeBCs"; then
	    if "$OptimizeOH"; then
		bcThreshold=$(($nElements - 5))
            else
		bcThreshold=$(($nElements - 4))
	    fi
            # The last four state vector elements are reserved for BC optimization of NSEW
            # domain edges. If the current state vector element is one of these, then
            # turn on BC optimization for the corresponding edge and revert emission perturbation
            if [ $x -gt $bcThreshold ]; then
		PerturbBCValues=$(generate_BC_perturb_values $bcThreshold $x $PerturbValueBCs)
		sed -i -e "s|CH4_boundary_condition_ppb_increase_NSEW:.*|CH4_boundary_condition_ppb_increase_NSEW: ${PerturbBCValues}|g" \
                       -e "s|perturb_CH4_boundary_conditions: false|perturb_CH4_boundary_conditions: true|g" \
                       -e "s|emission_perturbation_factor: ${PerturbValue}|emission_perturbation_factor: 1.0|g" \
                       -e "s|state_vector_element_number: ${xUSE}|state_vector_element_number: 0|g" geoschem_config.yml
            fi
	fi

	if "$OptimizeOH"; then
	    # The last state vector element is reserved for OH optimization.
            # If this is the current state vector element, then modify the OH
            # perturb value in HEMCO_Config.rc and revert emission perturbation.
	    OHthreshold=$(($nElements - 1))
            if [ $x -gt $OHthreshold ]; then
		sed -i -e "s|emission_perturbation: ${PerturbValue}|emission_perturbation: 1.0|g" \
                       -e "s|state_vector_element_number: ${xUSE}|state_vector_element_number: 0|g" geoschem_config.yml
		sed -i -e "s| OH_pert_factor  1.0| OH_pert_factor  ${PerturbValueOH}|g" HEMCO_Config.rc
            fi
	fi
```

The OH is global here. I was going to submit the hemispheric update separately but I can add that on top of these updates today just to get it in.

src/components/posterior_component/posterior.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+#SBATCH -o posterior.out
+#SBATCH -e posterior.err
```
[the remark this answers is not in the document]
msulprizio: In my posterior_runs directory, I do get a file slurm-[jobid].out. Those slurm files also get saved out in preview_runs, spinup_run, etc. I can't remember why I made this update so I will just remove those lines.

src/inversion_scripts/invert_calcJ.py
msulprizio: I'm not seeing this file called from anywhere, so I agree that it should be removed. If it is needed in the future, perhaps @megan-he can submit a separate pull request.

src/inversion_scripts/make_gridded_posterior.py
msulprizio: That was added by @megan-he in commit 24bac7b. I am fine with removing.

src/inversion_scripts/run_inversion.sh
msulprizio: Agreed. I think this was added while debugging, but should be removed.

src/inversion_scripts/run_inversion.sh
msulprizio: Good catch. This is now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43.

src/inversion_scripts/run_inversion.sh
msulprizio: Now removed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43.

src/inversion_scripts/jacobian.py
msulprizio: Now removed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43

src/inversion_scripts/invert_calcJ.py
msulprizio: Now removed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43.

src/components/posterior_component/posterior.sh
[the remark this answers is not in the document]
msulprizio: Now removed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43.

config.yml
```diff
@@ -72,18 +72,22 @@ StateVectorFile: "/home/ubuntu/integrated_methane_inversion/resources/statevecto
 ShapeFile: "/home/ubuntu/integrated_methane_inversion/resources/shapefiles/PermianBasin_Extent_201712.shp"
 
 ## Inversion
-## Note PriorError is a relative percentage 
-## and PriorErrorBCs is in ppb
+## Note PriorError and PriorErrorOH are relative percentages
```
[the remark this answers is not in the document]
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/c53f4d519860a678a5dec31f403770c497cfab43.

src/inversion_scripts/invert_calcJ.py
megan-he: Right this was just a temporary script, thanks!

run_imi.sh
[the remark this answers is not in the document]
laestrada: Can we do option 1? Thanks!

src/components/posterior_component/posterior.sh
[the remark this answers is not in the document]
laestrada: @msulprizio I think the slurm-[jobid].out files are generated from the ${RunName}_Posterior.run file which gets run from this function. That would be an update to `src/geoschem_run_scripts/ch4_run.template`

src/components/posterior_component/posterior.sh
```diff
@@ -162,7 +178,13 @@ run_posterior() {
...
     LatMinInvDomain=$(ncmin lat ${RunDirs}/StateVector.nc)
     LatMaxInvDomain=$(ncmax lat ${RunDirs}/StateVector.nc)
-    nElements=$(ncmax StateVector ${RunDirs}/StateVector.nc ${OptimizeBCs})
+    nElements=$(ncmax StateVector ${RunDirs}/StateVector.nc)
+    if "$OptimizeBCs"; then
+	nElements=$((nElements+4))
+    fi
+    if "$OptimizeOH";then
+	nElements=$((nElements+1))
+    fi
```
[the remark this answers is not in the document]
msulprizio: That's a good idea, but it would take some disentangling of the code so I suggest doing it as a separate pull request. 

Right now, we're utilizing this code in: 

1. Routine `setup_imi` within `src/components/setup_component/setup.sh`
2. Routine `setup_kf` within `src/components/kalman_component/kalman.sh`
3. Routine `reduce_dimension` within `components/statevector_component/statevector.sh`
4. Routine `run_posterior` within `components/posterior_component/posterior.sh`

The state vector file is currently created by calling `create_statevector` from within `setup_imi`. Moving the code for determining `nElements` out of `setup.sh` into `run_imi.sh` right after `setup_imi` is called would cause issues for routine `setup_jacobian` which uses `nElements` to determine how many run directories to create. One way around this is to remove the generation of the state vector file from `setup_imi` and into `run_imi.sh`. There may be other potential conflicts too, so we should think about how to do this more carefully.

src/inversion_scripts/invert.py
```diff
@@ -207,15 +215,24 @@ def do_inversion(
...
-        # add prior error for BCs
-        # as the last 4 elements of the diagonal
-        Sa_diag[-4:] = prior_err_bc**2
-        bc_idx -= 4
+
+    # If optimizing OH, adjust for it in the inversion
+    OH_idx = n_elements
+    if prior_err_oh > 0.0:
+        # add prior error for OH as the last element of the diagonal
+        Sa_diag[-1:] = (1/1000)*prior_err_oh**2 # try this
```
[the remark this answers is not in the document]
msulprizio: This was added by Megan commit ce1298e2. @megan-he do you remember why this was added?

src/inversion_scripts/invert.py
msulprizio: Your reasoning is correct here. I will also update the variable name for clarity.

OH is a scale factor so I have also updated the code and the few lines below to:

```
    # Update scale factors by 1 to match what GEOS-Chem expects
    # xhat = 1 + ratio
    # Notes:
    #  - If optimizing BCs, the last 4 elements are in concentration space,
    #    so we do not need to add 1
    #  - If optimizing OH, the last element also needs to be updated by 1
    xhat = ratio.copy()
    xhat[:scale_factor_idx] += 1
    if prior_err_oh > 0.0:
        xhat[n_elements] += 1
        print(f"xhat[OH] = {xhat[n_elements]}")
```

src/inversion_scripts/invert.py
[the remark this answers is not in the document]
megan-he: This was for weighing the OH term differently in the cost function in order to provide equal weight to OH (based on Bram's 2019 paper), for 1000 elements and 1 OH element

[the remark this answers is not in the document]
laestrada: @megan-he, that makes sense, so the 1000 should be a variable equal to (n_OH_elements/n_emission_elements)? Can we put Maasakers et al., 2019 in the comments to document this @msulprizio?

src/inversion_scripts/invert.py
laestrada: Great, thanks!

src/inversion_scripts/invert.py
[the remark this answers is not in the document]
megan-he: Yes!

run_imi.sh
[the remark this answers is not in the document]
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3.

src/inversion_scripts/imi_preview.py
```diff
@@ -526,9 +540,12 @@ def process(i):
     # Estimate information content
     # ----------------------------------
 
+    time_delta = enddate_np64 - startdate_np64
+    num_days = np.round((time_delta) / np.timedelta64(1, "D"))
+
     # State vector, observations
     emissions = np.array(emissions)
-    m = np.array(num_obs)  # Number of observations per state vector element
+    m = np.array(num_days)  # Number of observation days)
```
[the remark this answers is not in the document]
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3

src/inversion_scripts/invert.py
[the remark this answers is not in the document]
msulprizio: Now updated in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3.

src/inversion_scripts/invert.py
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3

src/inversion_scripts/invert.py
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3

src/inversion_scripts/invert.py
msulprizio: Now fixed in https://github.com/geoschem/integrated_methane_inversion/pull/157/commits/67f8f8a651ba6037f3cb882049755eb82ecb32c3

laestrada (approved):

PR #169, opened 2023-09-16
Feature/optimize domain edges

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This update implements boundary condition optimization as part of the inversion, making use of the new boundary condition perturbation options from @nicholasbalasus in geoschem 14.2.1.

This is done by effectively increasing the number of state vector elements by 4, and performing 4 additional perturbation simulations. Notably, we do the inversion calculations for these final 4 elements in concentration space instead of calculating a correction factor. The calculated increase/decrease in ppb is then applied in the posterior simulation.

These changes have been tested using our standard 1-week test case, but the differences are minimal when using buffer cells and BC optimization. Additional testing was done for 1-month over Tasmania to see the effect of having no buffer cells for the domain with and without BC optimization. Visualization notebooks are attached for this case.

In the tasmania case the posterior BCs are increased/decreased by `[-9.202384948730469, 9.279534339904785, 0.21053607761859894, -2.070218324661255]` ppb in NSEW domain edges. 

This reduces the bias in the posterior from -5.65 (in the inversion with no bc optimization) to -5.35 ppb when compared to the tropomi observations. It should be noted that there are only 300 observations in the domain, so we are unable to diverge too much from the prior.

[tasmania_1mon_bc_opt.html.txt](https://github.com/geoschem/integrated_methane_inversion/files/12664703/tasmania_1mon_bc_opt.html.txt)
[tasmania_1mon_no_bc_opt.html.txt](https://github.com/geoschem/integrated_methane_inversion/files/12664708/tasmania_1mon_no_bc_opt.html.txt)


src/inversion_scripts/calc_sensi.py
```diff
@@ -106,7 +109,10 @@ def process(h):
                 # Get the data for the current hour
                 pert = pert_data["SpeciesConcVV_CH4"][h, :, :, :]
                 # Compute and store the sensitivities
-                sensitivities = (pert.values - base.values) / perturbation
+                if (perturbationBC is not None) and (e > (nelements-4)):
+                    sensitivities = (pert.values - base.values) / (perturbationBC*1e-9)
```
laestrada, line 113: When optimizing BCs we need to convert the perturbation into ppb to match units with the values in the simulation, hence the `*1e-9`

nicholasbalasus:
Hi @laestrada  - I added a feature to make sure the GC perturbation is working as intended. Daniel V. would be better suited to review the inversion methodology.

[test-imi.txt](https://github.com/geoschem/integrated_methane_inversion/files/12674142/test-imi.txt) was run with `sbatch test-imi.sh` to test the feature I added. If we merge this into dev-2.0, we should keep in mind that perturbing boundary conditions won't work until we clone a version of GC with that feature available (I can make a PR for this once 14.2.1 is released).

If the test fails, it would fail into the .err file in the inversion folder. This doesn't crash the entire IMI (probably not desirable), but that's true for other errors here as well.

[visualization_notebook 2.html.zip](https://github.com/geoschem/integrated_methane_inversion/files/12675965/visualization_notebook.2.html.zip) has the results for the successful run.

src/components/jacobian_component/jacobian.sh
```diff
@@ -81,9 +82,26 @@ setup_jacobian() {
         fi
 	fi
    
-	# Update settings in geoschem_config.yml
-	sed -i -e "s|emission_perturbation: 1.0|emission_perturbation: ${PerturbValue}|g" \
-	       -e "s|state_vector_element_number: 0|state_vector_element_number: ${xUSE}|g" geoschem_config.yml
+	# Update settings in geoschem_config.yml except for the base run
+    if [ $x -ne 0 ]; then
+	    sed -i -e "s|emission_perturbation: 1.0|emission_perturbation: ${PerturbValue}|g" \
```
djvaron, line 87: It looks like the emission perturbation will be applied for any `x > 0`, including for the last four elements reserved for BC perturbations. Does it matter if the BC perturbation simulations also have an emission perturbation, or will that be ignored?

src/components/posterior_component/posterior.sh
```diff
@@ -100,6 +100,22 @@ run_posterior() {
         source ${GEOSChemEnv}
     fi
 
+    if "$OptimizeBCs"; then
+        if "$KalmanMode"; then
+            period_i=$((period_i+1))
```
djvaron, line 105: Maybe it's obvious but why are we doing `period_i + 1` ?

src/components/posterior_component/posterior.sh
```diff
@@ -100,6 +100,22 @@ run_posterior() {
...
+            inv_result_path="${RunDirs}/kf_inversions/period${period_i}/inversion_result.nc"
+        else
+            inv_result_path="${RunDirs}/inversion/inversion_result.nc"
+        fi
+        # set BC optimal delta values
+        PerturbBCValues=$(generate_optimized_BC_values $inv_result_path)
+        # turn on BC optimization for the corresponding edge and revert emission perturbation
```
djvaron, line 112: Minor thing -- this comment says "and revert emission perturbation" but it doesn't look like that happens or is necessary.

src/inversion_scripts/calc_sensi.py
```diff
@@ -106,7 +129,12 @@ def process(h):
                 # Get the data for the current hour
                 pert = pert_data["SpeciesConcVV_CH4"][h, :, :, :]
                 # Compute and store the sensitivities
-                sensitivities = (pert.values - base.values) / perturbation
+                if (perturbationBC is not None) and (e >= (nelements-4)):
+                    sensitivities = (pert.values - base.values) / perturbationBC
+                    if h != 0: # because we take the first hour on the first day from spinup
```
djvaron, line 134: I'm probably missing something, but what about when `h == 0` on other days (besides the first one)? And why do we need to perform the `test_GC_output_for_BC_perturbations` check for every time step?

src/inversion_scripts/run_inversion.sh
```diff
@@ -103,14 +103,14 @@ printf "DONE -- postproc_diags.py\n\n"
...
 if ! "$PrecomputedJacobian"; then
-
-    # 50% perturbation
-    Perturbation=0.5
-
+    python_args=(calc_sensi.py $nElements $PerturbValue $StartDate $EndDate $JacobianRunsDir $RunName $sensiCache)
+    # add an argument to calc_sensi.py if optimizing BCs
+    if "$OptimizeBCs"; then
+        python_args+=($PerturbValueBCs)
+    fi
```
djvaron, line 110: This is cool. Just checking, does it add the necessary space between the last argument `$sensiCache` and the new one `$PerturbValuesBCs` ?

djvaron: Same question below.

src/inversion_scripts/invert.py
```diff
@@ -253,6 +272,7 @@ def calculate_superobservation_error(sO, p):
     gamma = float(sys.argv[10])
     res = sys.argv[11]
     jacobian_sf = sys.argv[12]
+    prior_err_BC = float(sys.argv[13]) if len(sys.argv) > 13 else None
```
djvaron, line 275: I might be forgetting how `sys.argv` works, but just want to confirm the `>` condition is correct -- or should it be `>=` or `==` ?

src/components/posterior_component/posterior.sh
laestrada: Yeah this is definitely a bug! Not sure why I put this in -- Nice catch!

src/inversion_scripts/calc_sensi.py
laestrada: This just ensures that the BC perturbation in GEOS-Chem is behaving correctly to make sure, for example, someone isn't using an incompatible version of GEOS-Chem. @nicholasbalasus put this check in so I will defer to him, but.... My suspicion is that we don't necessarily need to do the check for every time step, and so skipping `h ==0` shouldn't matter for other days, as long as the other hours are correct. It's a pretty simple check so computationally shouldn't matter, but I suppose we could just do a spot check for each day eg. `if h == 1`

src/inversion_scripts/run_inversion.sh
laestrada: Yes, the reason being that `python_args` is now an array, so this actually just appends `$PertubValueBCs` to the array, then it gets unpacked for the actual python call (spaces automatically added).

src/inversion_scripts/invert.py
laestrada: I believe that `>` is correct because `invert.py` is called either like:
`python invert.py $nElements $JacobianDir $posteriorSF $LonMinInvDomain $LonMaxInvDomain $LatMinInvDomain $LatMaxInvDomain $PriorError $ObsError $Gamma $Res $jacobian_sf`

(in which `len(sys.argv)` would be exactly 13)
or:

`python invert.py $nElements $JacobianDir $posteriorSF $LonMinInvDomain $LonMaxInvDomain $LatMinInvDomain $LatMaxInvDomain $PriorError $ObsError $Gamma $Res $jacobian_sf $PriorErrorBCs`
(in which `len(sys.argv)` would be 14)

The thing to remember is that `sys.argv[0]` is always the name of the python file (here `invert.py`).

src/inversion_scripts/calc_sensi.py
nicholasbalasus: Either should be fine. The more checks the better (it should, but in case the perturbation doesn't work during some hours), but performance should be considered as well.

PR #172, opened 2023-10-12
update to GC 14.2.1

nicholasbalasus:

Name: Nick Balasus

### Describe the update

Update to GC 14.2.1. Note that the default test case has less DOFs now, but I think that is because the prior is very different now. 
[visualization_notebook 3.html.zip](https://github.com/geoschem/integrated_methane_inversion/files/12885966/visualization_notebook.3.html.zip)


msulprizio (approved):
Looks great. Thanks, @nicholasbalasus!

laestrada (approved):
Looks good!

PR #175, opened 2023-11-08
Feature: Lognormal error specification

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Implementation of lognormal prior error variance specification for iterative calculation of the nonlinear optimization problem through the Levenberg–Marquardt method (Rodgers, 2000). This adds two new config variables: `LognormalErrors`, a boolean for whether to use lognormal errors for the inversion, and `PriorErrorBufferElements`, which specifies the prior error in relative percentage for buffer elements. The latter is necessary because we only impose lognormal errors for the domain of interest. Buffer elements and optimized boundary condition elements continue to use normal error specification.

The implementation provides a separate script from `lognormal_invert.py` script to calculate xhat. Notably, rather than reading in the jacobian piece by piece and performing the inversion (like in `invert.py`), we read in the full jacobian. This significantly eases the implementation due to the iteration needed to solve for xhat.

Additionally, when using lognormal errors we add in a new "background" jacobian simulation with no emissions specified. This allows us to get a clean background and simply do `F(xhat) = ybkg+ K(xhat)` rather than needing to recalculate `F(xhat) ` for each iteration of `xhat`.

TODOs:
- change back config.harvard-cannon.yml file to default values
- Remove sed hack for background simulation once geoschem emissions toggle is fixed


config.yml
```diff
@@ -74,8 +74,11 @@ ShapeFile: "/home/ubuntu/integrated_methane_inversion/resources/shapefiles/Permi
 ## Inversion
 ## Note PriorError is a relative percentage
```
djvaron, line 75: Shouldn't call it a percentage but a fraction, I think. Since 0.5 doesn't mean 0.5%. Maybe something like this:

## Note PriorError is expressed as a fraction (e.g., 0.5 = 50%)

src/components/inversion_component/inversion.sh
```diff
@@ -14,19 +14,23 @@ setup_inversion() {
     cd ${OutputPath}/$RunName
     mkdir -p -v inversion
     mkdir -p inversion/data_converted
+    mkdir -p inversion/data_converted_background
     mkdir -p inversion/data_geoschem
+    mkdir -p inversion/data_geoschem_background
```
djvaron, line 19: Do we want to create `inversion/data_converted_background` and `inversion/data_geoschem_background` even when `LognormalErrors: false` ?

djvaron: I think users would generally want to create a new inversion directory when switching from normal to lognormal errors (for example as part of an inversion ensemble)

src/components/jacobian_component/jacobian.sh
```diff
@@ -41,21 +44,37 @@ setup_jacobian() {
...
+        x="background"
+        xstr=$x
+        create_simulation_dir
+    fi
+
+    printf "\n=== DONE CREATING JACOBIAN RUN DIRECTORIES ===\n"
+}
 
-	# Define the run directory name
+create_simulation_dir() {
```
djvaron, line 76: Nice simplification of the loop above.

src/components/jacobian_component/jacobian.sh
```diff
@@ -178,12 +213,23 @@ run_jacobian() {
...
+            precomputedJacobianCachePrefix=${ReferenceRunDir}/kf_inversions/period${period_i}
         else
             cd ${RunDirs}/inversion
-            precomputedJacobianCache=${ReferenceRunDir}/inversion/data_converted
+            precomputedJacobianCachePrefix=${ReferenceRunDir}/inversion
         fi
-        ln -s $precomputedJacobianCache data_converted_reference
+
+        precomputedJacobianCacheNormal=${precomputedJacobianCachePrefix}/data_converted
+        precomputedJacobianCacheLognormal=${precomputedJacobianCachePrefix}/data_converted_background
```
djvaron, line 223: is the `date_converted_background` directory where the Jacobian matrix elements are saved in the Lognormal errors case? If so, any reason why? Just checking.

src/inversion_scripts/invert.py
```diff
@@ -150,13 +150,7 @@ def do_inversion(
         obs_error = [obs if obs > 0 else 1 for obs in obs_error]
 
         # Jacobian entries for observations within bounds [ppb]
-        if jacobian_sf is None:
```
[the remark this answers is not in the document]
djvaron: I think I understand -- you added a logical to `run_inversion.sh` so that the Jacobian dir gets defined properly (precomputed or not) there. Right?

src/inversion_scripts/merge_partial_k.py
```diff
@@ -0,0 +1,94 @@
+import os
+import sys
+import yaml
+import pickle as pickle
+import numpy as np
+import xarray as xr
+from src.inversion_scripts.utils import load_obj, calculate_superobservation_error
+
+
+def merge_partial_k(satdat_dir, lat_bounds, lon_bounds, obs_err, background=False):
```
djvaron, line 10: This function would probably benefit from a long-form description / comment block.

src/inversion_scripts/lognormal_invert.py
```diff
@@ -0,0 +1,295 @@
+# Description: Script to perform inversion using lognormal errors
+# Usage: python lognormal_invert.py <path_to_config_file> <path_to_state_vector_file>
```
djvaron, line 2: There's also the `jacobian_sf` input variable

src/inversion_scripts/lognormal_invert.py
```diff
@@ -0,0 +1,295 @@
...
+from scipy.sparse import spdiags
+
+
+def lognormal_invert(config, state_vector_filepath, jacobian_sf):
+    """
+    Description:
+        Run inversion using lognormal errors following method from eqn 2 of
+        Chen et al., 2022 https://doi.org/10.5194/acp-22-10809-2022
+        Outputs inversion results to netcdf files.
+    Arguments:
```
djvaron, line 24: Arguments out of order and missing `jacobian_sf`

src/inversion_scripts/lognormal_invert.py
```diff
@@ -0,0 +1,295 @@
...
+    )
+    num_buffer_elems = int(config["nBufferClusters"])
+    num_normal_elems = num_buffer_elems + 4 if optimize_bcs else num_buffer_elems
+    ds = np.load("full_jacobian_K.npz")
+    K_temp = np.array(ds["K"]) * 1e9
+
+    # Apply scaling matrix if using precomputed Jacobian
+    if jacobian_sf is not None:
+        scale_factors = np.load(jacobian_sf)
+        # apply unit scaling for BC elements if optimizing BCs
```
djvaron, line 59: Is it possible for `jacobian_sf` to include elements for the BCs if the BCs were optimized in the previous inversion? Would the code here then append an additional (unneeded) four elements?

src/inversion_scripts/lognormal_invert.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # Calculate the difference between tropomi and the background
+    # simulation, which has no emissions
+    y_ybkg_diff = y - ybkg
+    ybkg, y, y_ybkg_diff = (
+        np.swapaxes(ybkg, 0, 1),
+        np.swapaxes(y, 0, 1),
+        np.swapaxes(y_ybkg_diff, 0, 1),
+    )
+    K_full = np.concatenate((K_ROI, K_normal), axis=1)
```
djvaron, line 83: Should this be moved up to line 70, after `K_ROI` and `K_normal` are defined?

src/inversion_scripts/lognormal_invert.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # fixed kappa of 10 following Chen et al., 2022 https://doi.org/10.5194/acp-22-10809-2022
+    kappa = 10
+    m, n = np.shape(K_ROI)
+
+    # Create base xa and lnxa matrices
+    # Note: the resulting xa matrix has lognormal elements until the final bc elements
```
djvaron, line 90: Should this say `the resulting xa vector has...` rather than `matrix`?

src/inversion_scripts/lognormal_invert.py
laestrada: This won't happen because `jacobian_sf` never has elements saved to it for the BCs when it is archived. This is because the way that we update the BCs in the posterior simulation is by updating `CH4_boundary_condition_ppb_increase_NSEW` directly in the HEMCO config file, rather than passing a file with the corrections in it like for emission and buffer elements. 

This does bring up a good point though, currently the BC optimization implementation resets the BC elements after each iteration (using the KF mode), and we are not iteratively passing in `CH4_boundary_condition_ppb_increase_NSEW` for successive prior simulations. Do you think we should be?

src/components/inversion_component/inversion.sh
laestrada: I initially did this to make it easier to test normal vs lognormal since the precomputed  jacobian option is currently broken when not using KF mode. I think you are right that we can just use a single directory for both.

One question is that under the current implementation if one ran a normal inversion and then wanted to do a lognormal inversion they would need to manually create and run a background simulation, as that simulation is not used in the normal inversion, but perhaps we should have it run that additional simulation by default, even in the normal case, to prevent the need for manual generation? Starting lognormal and switching to normal should work fine as is.

src/components/jacobian_component/jacobian.sh
laestrada: Same reason as above for the `data_geoschem_background` dir. Will work on merging them

src/inversion_scripts/lognormal_invert.py
djvaron: Let's keep the current reset approach for both buffer emission elements and NSEW BC elements of the state vector (as discussed earlier).

src/components/inversion_component/inversion.sh
djvaron: Another option that would be a bit more work but maybe more user friendly is to add a logical along these lines:

```
if $LognormalErrors is true and $PrecomputedJacobian is true:
    if the background simulation has not been run:
        run the background simulation
```

That way users running normal inversions wouldn't need to run an extra unnecessary simulation.

What do you think?

laestrada:
After fairly extensive testing of interaction with lognormal errors on OH, BC optimization, and global inversions from @eastjames and myself, we are merging this branch to dev

Issue #181, opened 2023-12-28
Global native resolution state vector creation issue


eastjames:
### Name and Institution (Required)

Name: James East
Institution: Harvard ACMG

### Description of your issue or question

I'm using the `feature/global_inversion` branch. When creating a native resolution state vector on the 2x2.5° grid, IMI creates a state vector element for all grid cells north of 60°S, totaling 10,800 state vector elements. A map looks like this:

```python
import xarray as xr
import matplotlib.pyplot as plt
import pycno

with xr.open_dataset('StateVector_orig.nc') as ds:
    sv_orig = ds['StateVector']

cno = pycno.cno(xlim=(-180,180),ylim=(-90,90))

sv_orig.plot()
cno.draw()
```
<img width="584" alt="Screenshot 2023-12-28 at 11 26 39 AM" src="https://github.com/geoschem/integrated_methane_inversion/assets/47705516/8afa9c45-b366-4fb3-835b-b99072dbaa45">

This happens in `src/utilities/make_state_vector_file.py`. The latitude of the HEMCO file used to filter for offshore O&G emissions gets manually adjusted, probably for regional applications. Also, the check for a land threshold always evaluates to `false` probably not intentionally.

Adding a check for mismatched longitudes at line 88-90 and changing the land_threshold check at line 139 seems to work:
```python
    # Require hemco diags on same global grid as land cover map
    if np.abs(lc.lon.values - hd.lon.values).max() != 0: #JDE add a check
        hd["lon"] = hd["lon"] - 0.03125  # initially offset by 0.03125 degrees
```

```python
    #if land_threshold: #old
    if land_threshold > 0: #JDE change
```

New state vector file:
<img width="577" alt="Screenshot 2023-12-28 at 11 26 50 AM" src="https://github.com/geoschem/integrated_methane_inversion/assets/47705516/a30afb0d-e9bc-44c6-be7e-85a4b93691f4">

nicholasbalasus:
Hey James, this was due to an issue in HEMCO that we [fixed](https://github.com/geoschem/HEMCO/issues/228). See [this](https://github.com/geoschem/integrated_methane_inversion/issues/141) IMI issue. The solution is to update the HEMCO standalone outputs used for this check using the updated version of HEMCO, which we should probably do before IMI 2.0 anyway given the emissions that have changed recently.

eastjames:
[...]

PR #182, opened 2024-01-05
Fix bugs in make_state_vector.py

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update

This PR addresses the following bugs in make_state_vector.py:

- an offset was mistakenly being applied in non 0.25 degree resolution grids
- faulty if statement caused land threshold to always evaluated to false
- In some cases labels would not be applied to buffer cells, causing gaps in the statevector labels resulting in perturbation simulations that would not be perturbed at all. This seems to only occur when using custom state vectors from a shapefile

To fix the 3rd bug I updated the buffer cell creation code and aligned them in `src/utilities/make_state_vector_file.py` and `src/notebooks/statevector_from_shapefile.ipynb`

@eastjames, tagging you and issue #181

### DIFF
--- a/src/notebooks/statevector_from_shapefile.ipynb
+++ b/src/notebooks/statevector_from_shapefile.ipynb
@@ -44,7 +44,8 @@
     "import matplotlib.pyplot as plt\n",
     "import yaml\n",
     "import warnings; warnings.filterwarnings(action='ignore')\n",
-    "sys.path.append('/home/ubuntu/integrated_methane_inversion/')\n",
+    "sys.path.append('../../')\n",
+    "from src.utilities.make_state_vector_file import cluster_buffer_elements\n",
     "from src.utilities.utils import download_landcover_files\n",
     "%matplotlib inline"
    ]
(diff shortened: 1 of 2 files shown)


eastjames:
Thanks, Lucas. Just a couple questions on the first 2 items.

For bullet #1 it looks good. @nicholasbalasus said in #181 that updating the HEMCO standalone outputs would get rid of the offset issue completely, so this may not be needed after that happens.

For bullet #2, this assumes that `land_threshold` is always set to a numerical value. Is that the expected behavior?

laestrada:
Hi @eastjamesm,

1. Yes, once the SA output is updated we can remove that line completely. 
2. Yes, `land_threshold` is set by `LandThreshold` in the config file and will always be a value between 0 and 1

eastjames:
Thanks, good to go.

nicholasbalasus:
Hey Lucas, if I remember correctly the 0.5° x 0.625° output is offset by 0.0625°.

nicholasbalasus:
This is untested and might not work with floating point issues, but maybe:
```
if np.abs(lc.lon.values - hd.lon.values).max() == 0.03125:
         hd["lon"] = hd["lon"] - 0.03125
elif np.abs(lc.lon.values - hd.lon.values).max() == 0.0625:
         hd["lon"] = hd["lon"] - 0.0625
```

eastjames:
Thanks Nick, assuming the difference is 0.03125 isn't flexible for other grids. What about this for flexibility to avoid enumerating each case, or are those the only two cases? Also untested for floating point issues.
```python
diff_threshold = 0.001
lon_diff = np.abs(lc.lon.values - hd.lon.values).max()
if lon_diff > diff_threshold:
	hd["lon"] = hd["lon"] - lon_diff
```

nicholasbalasus:
Looks good to me!

laestrada:
@eastjames and @nicholasbalasus,

I just did a quick check if the above would work and it looks like the IMI uses the nested-region specific landcover file, but the global file for emissions, so `np.abs(lc.lon.values - hd.lon.values).max()` yields an error due to differences in shapes. We could update it based on the bounds of landcoverfile, but given that we are planning to update the SA output files soon anyways I am going to update the lines to:

```
if np.abs(lc.lon.values - hd.lon.values).max() == 0.03125:
         hd["lon"] = hd["lon"] - 0.03125
elif np.abs(lc.lon.values - hd.lon.values).max() == 0.0625:
         hd["lon"] = hd["lon"] - 0.0625
```

and add a TODO reminding us to remove it when the diagnostics file is next updated

eastjames:
Thanks Lucas. I tested with global so I didn’t catch the error with different array shapes

PR #184, opened 2024-02-06
Feature: cleanup script

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Simple interactive utility script for cleanup of IMI output. Currently set up to delete:
- `OutputDir` files for jacobian simulations except for prior and background simulation
- `data_sensitivities` files in either normal or KF mode

I also include a dry-run mode that prints all the directories that will have recursive deletions without performing the removals.

Usage is:
`./cleanup_script.sh <config_file_path.yml>`

@djvaron @msulprizio @nicholasbalasus @eastjames please let me know if you think of additional files/directories that should be optionally deleted with this script

### DIFF
--- a/src/components/inversion_component/inversion.sh
+++ b/src/components/inversion_component/inversion.sh
@@ -29,6 +29,7 @@ setup_inversion() {
     cp ${InversionPath}/src/inversion_scripts/utils.py inversion/
     cp ${InversionPath}/src/inversion_scripts/run_inversion.sh inversion/
     cp ${InversionPath}/src/notebooks/visualization_notebook.ipynb inversion/
+    cp ${InversionPath}/src/utilities/cleanup_script.sh .
     sed -i -e "s:{INVERSION_PATH}:${InversionPath}:g" \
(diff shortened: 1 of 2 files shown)


msulprizio:
Thanks @laestrada! This is great. Do you think it would be worth copying this script to the IMI output directory so that users could easily access it without having to search for it in the IMI repo?

laestrada:
> Thanks @laestrada! This is great. Do you think it would be worth copying this script to the IMI output directory so that users could easily access it without having to search for it in the IMI repo?

Thanks Melissa, I think that is a good idea. I will make some updates, such that if no config file is provided it assumes the current directory is the output directory.

eastjames:
My only suggestion is to change the wording of the warning from "This is recommended if..." to something like "This data is generally no longer needed if..." Otherwise, looks good!

eastjames (approved):
My only recommendation is to change the warnings from "This is recommended if..." to something along the lines of "This data is generally no longer needed if..." Otherwise, looks good to me!

PR #185, opened 2024-02-07
conda activate bugfix

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Depending on the conda/miniconda/mamba distribution the way we load conda is faulty. It's best to source the bashrc file directly imo. This also makes it compatible with micromamba aliased as conda.

### DIFF
--- a/run_imi.sh
+++ b/run_imi.sh
@@ -50,8 +50,8 @@ eval $(parse_yaml ${ConfigFile})
 if ! "$isAWS"; then
     # Activate Conda environment
     printf "\nActivating conda environment: ${CondaEnv}\n"
-    eval "$(conda shell.bash hook)"
-    conda activate $CondaEnv
+    source ~/.bashrc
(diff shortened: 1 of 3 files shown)


msulprizio (approved):
Looks good to merge!

lizziel:
I had to restore the config file to the old way in order to work on a new cluster I am implementing the IMI on. I can make a PR to add the old way as a commented out option if the default way does not work.

laestrada:
Hi @lizziel 

We updated this to source a user-defined `CondaFile` in config.yml with the default set to `~/.bashrc`. I think it would work to make a conda_activate.sh script that looks like this:
```
#!/bin/bash
eval "$(conda shell.bash hook)"
```

and then set `CondaFile: /path/to/conda_activate.sh`

What do you think?

lizziel:
Great idea! I tested this and it works.

PR #193, opened 2024-03-08
Updates to speedup Jacobian runtime

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

Two major updates have been made to help speed up the Jacobian runs within the IMI. 

(1) A prior component has been added and is activated by setting `DoPriorEmis: true` in config.yml. This will run a HEMCO standalone simulation to generate a prior emissions output (netCDF file). This netCDF file is later used in the jacobian simulations to read in total emissions rather than individual emissions inventories and sectors, which greatly reduce I/O and speeds up simulations by 50%.

(2) An option has been added to combine jacobian runs into fewer runs by adding state vector elements as individual CH4 tracers. In the current IMI, a GEOS-Chem simulation is run for each state vector element, perturbing the emissions for that element. With the `NumJacobianRuns` setting users can specify the total number of jacobian runs they would like to submit and the IMI will determine how many state vector elements should be included in each run based on that number. Rather than being represented by separate run directories, each state vector element is represented by a CH4 tracer in GEOS-Chem (e.g. CH4_0000, CH4_0001). Perturbations will be applied to each state vector element using the StateVector.nc file and scale factors defined in HEMCO_Config.rc. 

The default value for `NumJacobianRuns` is 10. This is because initial testing on the Harvard Cannon cluster has found that ~10 state vector elements per run is ideal (see figure below). This may vary by system.

Screenshot 2024-03-08 at 10 38 52 AM

To turn off combining jacobian runs, users can specify `NumJacobianRuns: -1` in which case a simulation will be run for each state vector element.


config.yml
```diff
@@ -111,6 +112,7 @@ DoPreview: true
 DOFSThreshold: 0
 
 ## Resource allocation settings for slurm jobs
+PriorMemory: 50000
```
laestrada, line 115: A few things: 
1) Perhaps the default should be set to 32000 to be consistent with `SimulationMemory`. Do you think we should have `PriorEmis` use the configured `SimulationMemory` or is there benefit to keeping these variables separate? 

2) Should there also be `CombineJacobianRuns` somewhere in config.yml or is it assumed false if not present? I dont actually see this variable referenced anywhere.. is this used?

3) Can you add new config variables in sanitize_input_yaml.py? This would also be a good place to enforce that if combining jacobian run, UseTotalPriorEmis is true.

docs/source/getting-started/imi-config-file.rst
```diff
@@ -205,8 +205,9 @@ These settings turn on/off (``true`` / ``false``) different steps for running th
    :widths: 30, 70
    :class: tight-table
 
-   * - ``RunSetup``
-     - Boolean to run the setup script (``setup_imi.sh``), including selected setup modules above.
+   * - ``DoPriorEmis``
```
laestrada, line 208: Does this generate the prior emissions for the entire period? Does this work with kalman mode?

envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
```diff
@@ -63,7 +63,7 @@ PointSourceDatasets: ["SRON"]
 ReducedDimensionStateVector: false
 DynamicKFClustering: false
 ClusteringMethod: "kmeans"
-NumberOfElements: 45
+NumberOfElements: 1000
```
laestrada, line 66: I think this is greater than the number of native state vector elements in the default configuration, which could be confusing for a user if `ReducedDimensionStateVector` is toggled on in a test run. Switch back to 45

docs/source/getting-started/imi-config-file.rst
```diff
@@ -247,11 +248,18 @@ Note: some python scripts are also deployed using slurm and default to using the
...
    * - ``JacobianMemory``
      - Amount of memory to allocate to each jacobian simulation (in MB).
+   * - ``RequestedTime``
+     - Max amount of time to allocate to each sbatch job (eg. "0-6:00")
    * - ``SchedulerPartition``
      - Name of the partition(s) you would like all slurm jobs to run on (eg. "debug,huce_intel,seas_compute,etc").
    * - ``MaxSimultaneousRuns``
      - The maximum number of jacobian simulations to run simultaneously. The default is -1 (no limit) which will submit all jacobian simulations at once. If the value is greater than zero, the sbatch array statement will be modified to include the "%" separator and will limit the number of simultaneously running tasks from the job array to the specifed value.
- 
+   * - ``NumJacobianRuns``
```
laestrada, line 257: Add documentation for `CombineJacobianRuns` too?

src/components/jacobian_component/jacobian.sh
```diff
@@ -44,12 +55,21 @@ setup_jacobian() {
...
            -e "s:{InversionPath}:${InversionPath}:g" jacobian_runs/run_prior_simulation.sh
 
+    # Modify HEMCO_Config.rc to turn off individual emission inventories
+    # and use total emissions saved out from prior emissions simulation
+    # instead
+    # Do this in template run directory to avoid having to repeat for each
+    # Jacobian run directory
+    printf "\nTurning on use of total prior emissions in HEMCO_Config.rc.\n"
+    sed -i -e "s|UseTotalPriorEmis      :       false|UseTotalPriorEmis      :       true|g" \
+           -e "s|AnalyticalInversion    :       false|AnalyticalInversion    :       true|g" \
```
laestrada, line 65: Was use analytical inversion false before?

src/components/jacobian_component/jacobian.sh
```diff
@@ -44,12 +55,21 @@ setup_jacobian() {
...
+           -e "s|GFED                   : on|GFED                   : off|g" ${RunTemplate}/HEMCO_Config.rc
```
laestrada, line 66: I would think that setting `UseTotalPriorEmis` true would prevent GFED from being used anyways?

src/components/prior_component/prior.sh
```diff
@@ -0,0 +1,107 @@
...
+    -p $SchedulerPartition \
+    -W ${PriorName}.run; wait;
+
+    # check if exited with non-zero exit code
+    [ ! -f ".error_status_file.txt" ] || imi_failed $LINENO
+    
+    printf "\nDone prior emissions simulation\n\n"
+    
+    printf "\n=== DONE GENERATING PRIOR EMISSIONS ===\n"
+    prior_end=$(date +%s)
```
laestrada, line 106: If a user wanted to modify their prior emissions would they still edit the template rundirectory, this run directory, or both?

laestrada (changes requested):
Thanks Melissa! This looks great -- a few small changes requested (see inline comments) and points of confusion that I would like to clear up.

config.yml
msulprizio: > 1. Perhaps the default should be set to 32000 to be consistent with `SimulationMemory`. Do you think we should have `PriorEmis` use the configured `SimulationMemory` or is there benefit to keeping these variables separate?

Yes, I think it makes sense for them to be consistent. I believe I had separated them under the impression that the HEMCO standalone would require drastically different memory allocation but they're close enough that they can use the same value.

> 2. Should there also be `CombineJacobianRuns` somewhere in config.yml or is it assumed false if not present? I dont actually see this variable referenced anywhere.. is this used?

No, the `CombineJacobianRuns` logical was removed during some of the cleanup. It's now determined simply from `NumJacobianRuns` -- if that value is >0 then the jacobian runs will be combined into the specified number. I've updated the description of this pull request to avoid confusion.

> 3. Can you add new config variables in sanitize_input_yaml.py? This would also be a good place to enforce that if combining jacobian run, UseTotalPriorEmis is true.

Both `CombineJacobianRuns` and `UseTotalPriorEmis` were removed in later commits of this PR during cleanup. The IMI will always use total prior emissions (unless a user goes in and manually changes that in `template_run/HEMCO_Config.rc` before setting up the jacobian runs). This was changed because it simplifies applying the emissions perturbations in HEMCO_Config.rc.

With that said, I have updates `sanitize_input_yaml.py` to include new fields added to the config file in this PR.

docs/source/getting-started/imi-config-file.rst
msulprizio: I set it up to run the prior (HEMCO standalone) simulation for 1 hour using the start date specified in the config file. However, it may make more sense to run for the entire period and save out the average emissions rates (default units are kg/m2/s). For example, using the emission rates for the start date doesn't necessarily make sense if running an annual inversion which includes seasonally varying emissions. 

I've made this change in commit 29f46830.

envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
msulprizio: The value was raised to 1000 in this config file only, which is meant to serve as a starting point for the global inversion. I selected this value based on approximations from Meghan He's work at 2x2.5. Having only 45 elements in the global inversion doesn't make much sense. 

The value remains 45 in both `config.yml` and `envs/Harvard-Cannon/config.harvard-cannon.yml`.

docs/source/getting-started/imi-config-file.rst
msulprizio: CombineJacobianRuns has been removed during cleanup. I fixes the description of this PR above to reflect that.

src/components/jacobian_component/jacobian.sh
msulprizio: `AnalyticalInversion` is false by default when setting up a GEOS-Chem run directory for CH4 simulations. It remained false in the IMI prior to these updates. This is because we were previously applying the emissions perturbations in the GEOS-Chem source code (GeosCore/global_ch4_mod.F90). That code has since been removed and we now apply emissions perturbations directly in HEMCO_Config.rc by setting `AnalyticalInversion: true`.

src/components/jacobian_component/jacobian.sh
msulprizio: Unfortunately, it does not because GFED is a HEMCO extension. We can use `UseTotalPriorEmis` in HEMCO_Config.rc to block off the emissions data read for GFED emissions, but that does not prevent HEMCO from calling the GFED extension which will then look for the (now missing) data. Therefore, we have to also set the GFED extension to off.

src/components/jacobian_component/jacobian.sh
```diff
@@ -139,13 +148,104 @@ setup_jacobian() {
...
+	# For all other runs, just disable Restarts
 	else
             if "$HourlyCH4"; then
 		sed -i -e 's/'\''Restart/#'\''Restart/g' HISTORY.rc
             fi
 	fi
 
+	# Determine start and end element numbers for this run directory
+	if [ $NumJacobianRuns -lt 0 ]; then
+	    start=$x
```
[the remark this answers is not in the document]
msulprizio: This is now fixed in commit 59301b55.

src/components/jacobian_component/jacobian.sh
```diff
@@ -139,13 +148,104 @@ setup_jacobian() {
...
+	    sed -i -e "/$HcoPrevLine4/a $HcoNewLine4" HEMCO_Config.rc
+	    HcoPrevLine4='BC_CH4_'$istr
+
+	    # Add new Perturbations.txt and update
+	    cp Perturbations.txt Perturbations_${istr}.txt
+	    PertNewLine='\
+ELEM_'$istr'  '$i'     '$PerturbValue''
+	    sed -i "/$PertPrevLine/a $PertNewLine" Perturbations_${istr}.txt
+
+	done
```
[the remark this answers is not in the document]
msulprizio: Correct. Emission perturbations are now done exclusively by applying scale factors in HEMCO_Config.rc. Those scale factors are applied to the specified state vector element (grid cells or clusters) obtained from StateVector.nc.

For boundary conditions and OH, those perturbations are also applied in HEMCO_Config.rc. We create additional run directories for each of those perturbations. In those HEMCO_Config.rc files, `AnalyticalInversion` is set to false. In the case of OH, the value for `OH_pert_factor` in HEMCO_Config.rc:

```
2 OH_pert_factor  1.0 - - - xy 1 1
```

is modified from 1.0 to `PerturbValueOH`: 
 https://github.com/geoschem/integrated_methane_inversion/blob/5729e7523307ba040eca3c9021807a1b50af6e15/src/components/jacobian_component/jacobian.sh#L124).

For BCs, the perturbation still occurs within the GEOS-Chem source code (in [GeosCore/set_boundary_conditions.F90](https://github.com/geoschem/geos-chem/blob/20e2402baf56c682cc04af74adb139efdb6ca000/GeosCore/set_boundary_conditions_mod.F90#L156-L199)) when BCs are applied to the species concentration array  as implemented by @nicholasbalasus.

src/components/prior_component/prior.sh
msulprizio: If a user wanted to modify their prior emissions, they could do so in the `prior_run` directory then rerun the IMI with `DoPriorEmis: true`. They could also modify it in the `template_run` directory and rerun with `DoPriorEmis: true` (make sure to either remove their previous `prior_run` directory or set `SafeMode: false` in the config file. In either case, the total CH4 emissions saved out from the prior run will be used in the subsequent jacobian runs.

src/components/statevector_component/aggregation.py
```diff
@@ -224,14 +224,16 @@ def get_max_aggregation_level(config, sensitivities, desired_element_num):
...
         max_aggregation_level = 64
-    elif config["Res"] == "0.25x0.3125":
-        max_aggregation_level = 256
+    elif config["Res"] == "2.0x2.5":
+        max_aggregation_level = 16
+    elif config["Res"] == "4.0x5.0":
+        max_aggregation_level = 4
 
-    background_elements_needed = np.ceil(len(sensitivities) / max_aggregation_level)
+    background_elements_needed = np.ceil(len(sensitivities) / max_aggregation_level) # 1-month: 10800/4 = 2700 > 600
```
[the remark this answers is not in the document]
msulprizio: This was inadvertently brought in from early development of the global inversion capability. I have now removed it in commit 371d5ab6.

PR #196, opened 2024-03-23
Bugfix/run_inversion.sh error handling

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Previously if the run_inversion.sh script failed with an error, the inversion workflow would continue despite the error because run_inversion.sh is run via an sbatch job. 

To fix this, on any non-zero exit code this script writes out a hidden error file that the inversion workflow reads and errors out on. This is how we handle other failed sbatch jobs eg. jacobian runs in the IMI.


msulprizio (approved):
Looks good to merge!

PR #200, opened 2024-04-10
fix bug where symlinked restart would be overwritten by default restart

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This fixes a bug where the prior restart file, if downloaded using a gcclassic dryrun, would overwrite the existing symlinked restart file. We avoid this problem by removing the line for the restart from the log.dryrun file.


src/components/jacobian_component/jacobian.sh
```diff
@@ -47,125 +47,128 @@ setup_jacobian() {
...
+        if [ $x -eq 0 ]; then
             if "$ProductionDryRun"; then
-		printf "\nExecuting dry-run for production runs...\n"
-		./gcclassic --dryrun &> log.dryrun
-		./download_data.py log.dryrun aws
+                printf "\nExecuting dry-run for production runs...\n"
+                ./gcclassic --dryrun &>log.dryrun
+                # prevent restart file from getting downloaded since
+                # we don't want to overwrite the one we link to above
+                sed -i '/GEOSChem.Restart/d' log.dryrun
```
laestrada, line 162: @msulprizio this is the only actual change that is not just a change in whitespace.

msulprizio (approved):
Looks good to merge. Thanks @laestrada!

PR #208, opened 2024-05-04
bugfix accounting for clustering in superobservation calculation in p…

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This fixes a bug in the IMI preview where the number of superobservations was considered to be the number of observation days. This is a valid assumption in the case of tropomi, but only if using native resolution. If the state vector is clustered, the number of superobservations should be num_days * num_native_grid_cells. This bug was causing the estimated DOFS to be artificially very low because large clusters were only considered to have a single superobservation per day.


megan-he:
Looks good thanks Lucas!

Issue #210, opened 2024-05-06
[BUG] Soil sink in EmisCH4_Total and sectoral breakdown


djvaron:
Name: Daniel Varon
Institution: Harvard University

### Soil sink bug when computing total emissions, sectoral breakdown

The visualization notebook creates the posterior emissions from the prior emissions and scale factors:

```
# Prior emissions
prior = xr.load_dataset(prior_pth)["EmisCH4_Total"].isel(time=0)
if config["KalmanMode"]:
    prior_sf = xr.load_dataset(prior_sf_pth)
    prior = prior * prior_sf["ScaleFactor"]

# Optimized scale factors
scale = xr.load_dataset(results_pth)["ScaleFactor"]

# Posterior emissions
posterior = prior * scale
```

The soil absorption sink should be removed from `EmisCH4_Total` before applying posterior scale factors, because soil absorption is not currently optimized in the inversion. Not doing so leads to incorrect posterior emission output and sectoral breakdown.

This issue may affect other parts of the codebase where `EmisCH4_Total` is called, including the Kalman filter feature (e.g., `prepare_sf.py`)

The posterior simulation should NOT be affected because soil absorption is correctly removed before the scale factors are applied.

djvaron:
Here are potential fixes for the visualization notebook.

Defining the prior and posterior emissions:

```
# Prior emissions
hemco_emissions = xr.load_dataset(prior_pth).isel(time=0)
prior = hemco_emissions["EmisCH4_Total"].copy()
prior_noSoilAbsorb = prior - hemco_emissions["EmisCH4_SoilAbsorb"]

# Optimized scale factors
scale = xr.load_dataset(results_pth)["ScaleFactor"]

# Posterior emissions
posterior = prior_noSoilAbsorb * scale + hemco_emissions["EmisCH4_SoilAbsorb"]
```

Sectoral breakdown:

```
def my_autopct(
    pct,
):  # Only plots the data labels for sectoral emissions that are larger than 15%
    return ("%.1f%%" % pct) if pct > 15 else ""

# load in full sectors based on prior distribution
posterior_sectors = xr.load_dataset(prior_pth) * scale

# overwrite scaled soil absorption with original soil absorption
soil_absorb = xr.load_dataset(prior_pth)["EmisCH4_SoilAbsorb"]
posterior_sectors["EmisCH4_SoilAbsorb"] = soil_absorb

# extract sector names
[...]

PR #213, opened 2024-05-07
Bugfix/soil absorption accounting

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update

This bugfix addresses #210. Effectively, the soil absorption sink is not optimized in the inversion and we were not properly accounting for this when calculating the posterior emissions.


msulprizio (approved):
Thanks for these fixes. Looks good to me!

megan-he (approved):
Looks good, thanks Lucas!

src/components/kalman_component/prepare_sf.py
```diff
@@ -81,9 +82,11 @@ def prepare_sf(config_path, period_number, base_directory, nudge_factor):
...
 
             # Get posterior emissions multiplied up to current period p, and apply nudging
-            current_posterior_emis = (
-                posterior_p["ScaleFactor"] * sf["ScaleFactor"] * original_emis
-            )
+            # Account for non-optimized soil sink by calling get_posterior_emissions()
+            posterior_scale_ds["ScaleFactor"] = posterior_p["ScaleFactor"] * sf["ScaleFactor"]
+            posterior_emis_ds = get_posterior_emissions(original_emis_ds, posterior_scale_ds)
+            current_posterior_emis = posterior_emis_ds["EmisCH4_Total"].isel(time=0, drop=True)
+
```
laestrada, line 89: One point I am unsure of is whether we should be removing the soil sink in the calculation of the nudged scale factor below (line 108). I think the answer is yes, but would like to check what @megan-he, @msulprizio, and especially @djvaron think.

djvaron: This is a good question! I _think_ what we want to do is:

1. Remove soil sink from current posterior and original prior
2. Do weighted average of current posterior and original prior
3. Apply lambda scaling
4. Compute scale factors to apply in next period (i.e., sf["ScaleFactor"] on line 108)

src/components/kalman_component/prepare_sf.py
```diff
@@ -5,7 +5,19 @@
...
+    """
+    Remove soil absorption from total emissions and return the new total.
+    
+    Arguments
+        emis [xr.Dataset] : HEMCO emissions dataset
+    Returns
+        [xr.DataArray] : Total emission from all sources except soil absorption
+    """
+    ds = emis.copy()
+    ds["EmisCH4_Total"] = ds["EmisCH4_Total"] - ds["EmisCH4_Soil"]
```
djvaron, line 18: Shouldn't it be `EmisCH4_SoilAbsorb` ?

PR #221, opened 2024-05-17
Feature/upgrade conda env

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Upgrade conda environment to newest compatible packages. I have tested this for a 1-month full permian inversion and, with the updates, there is zero differences in the resulting inversion using the old/new environment.


eastjames:
Adding `  - dask=2024.6.2` to the environment file creates a working environment. Versions of other libraries are unchanged. I haven't tested it for changes to results.

PR #224, opened 2024-05-24
Bugfix/allow zero buffer elems

laestrada:

Name: Lucas Estrad
Institution: Harvard ACMG

### Describe the update
Previously the IMI required use of buffer elements in the state vector when making a rectilinear state vector. THis update allows specification of 0 buffer elements by setting the buffer region == -9999. Note that users should still leave enough buffer region to account for the geoschem "buffer zone". @djvaron @msulprizio how much buffer zone is required? Perhaps I should enforce at least that amount in `make_state_vector_file.py`?


msulprizio:
>how much buffer zone is required? 

GEOS-Chem nested-grid simulations use a default buffer zone of 3 grid cells on each edge, so the actual extent depends on the model resolution. We [tell users](https://geos-chem.readthedocs.io/en/stable/supplemental-guides/nested-grid-guide.html?highlight=buffer#i-noticed-abnormal-concentrations-at-boundaries-of-the-nested-region-is-that-normal) to exclude those boxes from their analysis.

laestrada:
Thanks @msulprizio! That is very helpful -- I will add some logic to add 3 grid boxes to each side based on the resolution

PR #226, opened 2024-06-05
Bugfix/jacobian perturbation fix

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Lee Murray identified an issue in our Jacobian construction where noise due to nonlinearities in advection can cause negative values in the Jacobian construction. @nicholasbalasus demonstrated that this can be fixed by running jacobian perturbation simulations with: low background of methane (1ppb), low initial conditions of methane (1ppb), and a high emissions perturbation (10e-8 kg/m2/s). 

Here we implement these changes into the IMI jacobian construction by applying variable scale factors to emissions, such that perturbations achieve the high perturbation level. 

This PR has had basic testing with a standalone inversion, lognormal errors, kalman mode, and BC/OH perturbations. Global inversions have not been tested yet.

There are also quite a few bugfixes for the jacobian speedup in this branch.


src/components/jacobian_component/jacobian.sh
```diff
@@ -151,52 +185,176 @@ create_simulation_dir() {
...
+            # Use nudged scale factors for the prior simulation and OH simulation for kalman mode
+            sed -i -e "s|--> Emis_PosteriorSF       :       false|--> Emis_PosteriorSF       :       true|g" \
+                -e "s|--> UseTotalPriorEmis      :       false|--> UseTotalPriorEmis      :       true|g" \
+                -e "s|gridded_posterior.nc|${RunDirs}/ScaleFactors.nc|g" HEMCO_Config.rc
+        fi
+
+    else
+        # set 1ppb CH4 boundary conditions and restarts for all other perturbation simulations
+        # Note that we use the timecycle flag C to avoid having to make additional files
+        RestartFile=${RunDirs}/jacobian_1ppb_ics_bcs/Restarts/GEOSChem.Restart.1ppb.${StartDate}_0000z.nc4
```
megan-he, line 252: If using BCs for the Restart (which I am doing for my global simulation), then the Restart File in `jacobian_1ppb_ics_bcs/Restarts/` is called `GEOSChem.**BoundaryConditions**.1ppb.${StartDate}_0000z.nc4` instead of `GEOSChem.**Restart**.1ppb.${StartDate}_0000z.nc4`, so I added logic to prevent a broken symlink from being created 10 lines down, like this:
[https://github.com/megan-he/integrated_methane_inversion/commit/52aebb0721edf9b036f1ee848da09c05517894a5](url)

laestrada: @megan-he I am guessing your fork of the IMI is private, so I cant see your fix :(

megan-he: Here is the correct link https://github.com/megan-he/integrated_methane_inversion/commit/52aebb0721edf9b036f1ee848da09c05517894a5

nicholasbalasus:
This branch didn't work after [4a46a9f](https://github.com/geoschem/integrated_methane_inversion/pull/226/commits/4a46a9f2cc446f51c9ea3eb3197fd4452f6f7e61) since `$UseBCsForRestart` is always true. [cef3c72](https://github.com/geoschem/integrated_methane_inversion/pull/226/commits/cef3c72a8bc2c23e0776ad45ad6ce6cdc0b8579e) should fix that.

nicholasbalasus:
Changes so far:

- turned off meteorology for the HEMCO standalone so it runs fast now
- set the default perturbation to be 1e-8 kg/m2/s instead of 1e-7
- stopped GFED emissions from running in the posterior
- removed outdated settings modifications to GC configuration files

nicholasbalasus:
Hi @laestrada, @msulprizio - congrats on the impressive and extensive work here :)

I've reviewed this and added some fixes, but mostly in the context of the test case in the default config file (subset of the Permian for 1 week). For this case, as of ecf6b99, this branch is working, as tested by this script:

```
#!/bin/bash

for branch in dev fix; do
    cd "${HOME}/for-lucas"
    mkdir "${branch}"
    cd "${branch}"
    git clone https://github.com/geoschem/integrated_methane_inversion.git
    cd integrated_methane_inversion
    if [ $branch = dev ]; then
        git checkout 06aa4c1
    elif [ $branch = fix ]; then
        git checkout ecf6b99
    fi
    sed -i -e "s|Test_IMI_Permian|Test_IMI_Permian_${branch}|g" \
        -e "s|BlendedTROPOMI: false|BlendedTROPOMI: true|g" \
        -e "s|SetupSpinupRun: false|SetupSpinupRun: true|g" \
        -e "s|SetupJacobianRuns: false|SetupJacobianRuns: true|g" \
        -e "s|SetupInversion: false|SetupInversion: true|g" \
        -e "s|SetupPosteriorRun: false|SetupPosteriorRun: true|g" \
        -e "s|DoSpinup: false|DoSpinup: true|g" \
        -e "s|^DoJacobian: false|DoJacobian: true|g" \
        -e "s|DoInversion: false|DoInversion: true|g" \
        -e "s|DoPosterior: false|DoPosterior: true|g" \
        -e "s|/tropomi|/blended|g" envs/Harvard-Cannon/config.harvard-cannon.yml
    sbatch -p huce_cascade -t 2-00:00 run_imi.sh envs/Harvard-Cannon/config.harvard-cannon.yml
done
```

That being said, I wouldn't be surprised if there are problems for the following configurations:
- global
- lognormal
- kalman filter
- non-kalman filter inversions that cross months (@sarahhancock, @eastjames - maybe check that the emissions are updating from month-to-month in your tests)

Given how large this PR has become, maybe it would be most efficient to merge this into `dev` as is and work through problems as they arise for the above cases as people test extensively from `dev`.

nicholasbalasus (approved):

Issue #206, opened 2024-05-01
Boundary condition files wrong size time dimension


eastjames:
Name: James East
Institution: Harvard ACMG

The `time` dimension on TROPOMI and blended boundary condition files on 2018-04-03 is the wrong size and HEMCO fails when reading these files as a result. This affects `v2024-01` and `v2024-03`.

Thanks to @laestrada for identifying the issue.

These files have 1 time step, but HEMCO expects 8. For example:
in
```bash
$ ncks -m -v time /n/holylfs05/LABS/jacob_lab/imi/ch4/blended-boundary-conditions/v2024-01/GEOSChem.BoundaryConditions.20180403_0000z.nc4
```
out
```
netcdf GEOSChem.BoundaryConditions.20180403_0000z {
  dimensions:
    time = UNLIMITED ; // (1 currently)
```

in
```bash
$ ncks -m -v time /n/holylfs05/LABS/jacob_lab/imi/ch4/blended-boundary-conditions/v2024-03/GEOSChem.BoundaryConditions.20180403_0000z.nc4
```
out
```
netcdf GEOSChem.BoundaryConditions.20180403_0000z {
  dimensions:
    time = UNLIMITED ; // (1 currently)
```

Should be:
in
```bash
ncks -m -v time /n/holylfs05/LABS/jacob_lab/imi/ch4/blended-boundary-conditions/v2024-03/GEOSChem.BoundaryConditions.20180404_0000z.nc4
```
out
```
netcdf GEOSChem.BoundaryConditions.20180404_0000z {
  dimensions:
    time = UNLIMITED ; // (8 currently)
```

in
```bash
ncks -m -v time /n/holylfs05/LABS/jacob_lab/imi/ch4/blended-boundary-conditions/v2024-01/GEOSChem.BoundaryConditions.20180404_0000z.nc4
```
out
```
netcdf GEOSChem.BoundaryConditions.20180404_0000z {
  dimensions:
    time = UNLIMITED ; // (8 currently)
```

nicholasbalasus:
Thanks, James. That's really weird. If it isn't urgent, I can address this when I make new BCs in the next week or two. Can you start with 20180430 instead? The 20180401-20180429 files have no TROPOMI data for correction so they're not very useful.

eastjames:
[...]

PR #227, opened 2024-06-21
Feature/v2024-06-bcs

nicholasbalasus:

Name: Nick Balasus
Institution: Harvard

### Describe the update

These are new boundary conditions using GC 14.4.0. I updated some of the smoothing methods, but the changes are minor.

#206 can be closed when this is merged.

A PR should follow this to update the IMI to use GC 14.4.0.


laestrada (approved):
Looks good -- thanks!

Issue #228, opened 2024-06-21
SRON plume retrieval issue


nicholasbalasus:
### Name and Institution (Required)

Name: Nick Balasus
Institution: Harvard University

### Description of your issue or question

Running the IMI dev branch (as of commit e18648e) errors on the plume visualization out of the box (20180501-20180508).

```
Using cached plume data for SRON
/n/home06/nbalasus/test-imi-dev/integrated_methane_inversion/src/inversion_scripts/plumes.py:391: UserWarning: Cannot open cached file at /n/holyscratch01/jacob_lab/nbalasus/Test_IMI_Permian_dev/SRON_plumes/plumes.geojson, fetching new data.
  warnings.warn(msg)
Fetching plumes from SRON...
Traceback (most recent call last):
  File "/var/slurmd/spool/slurmd/job37032535/slurm_script", line 679, in <module>
    imi_preview(
  File "/var/slurmd/spool/slurmd/job37032535/slurm_script", line 242, in imi_preview
    point_sources=get_point_source_coordinates(config),
  File "/n/home06/nbalasus/test-imi-dev/integrated_methane_inversion/src/inversion_scripts/point_sources.py", line 79, in get_point_source_coordinates
    observers.append( observer(config, usecached = True) )
  File "/n/home06/nbalasus/test-imi-dev/integrated_methane_inversion/src/inversion_scripts/plumes.py", line 450, in __init__
    super().__init__(self.myname, config, usecached, self.cache)
  File "/n/home06/nbalasus/test-imi-dev/integrated_methane_inversion/src/inversion_scripts/plumes.py", line 398, in __init__
    self._cache_data()
  File "/n/home06/nbalasus/test-imi-dev/integrated_methane_inversion/src/inversion_scripts/plumes.py", line 417, in _cache_data
    self.gdf.to_file(
  File "/n/home06/nbalasus/micromamba/envs/imi_env/lib/python3.9/site-packages/geopandas/geodataframe.py", line 1114, in to_file
    _to_file(self, filename, driver, schema, index, **kwargs)
  File "/n/home06/nbalasus/micromamba/envs/imi_env/lib/python3.9/site-packages/geopandas/io/file.py", line 367, in _to_file
    schema = infer_schema(df)
  File "/n/home06/nbalasus/micromamba/envs/imi_env/lib/python3.9/site-packages/geopandas/io/file.py", line 428, in infer_schema
    raise ValueError("Cannot write empty DataFrame to file.")
[...]

PR #229, opened 2024-06-27
Fix logic for case when no plumes available

eastjames:

Name: James East
Institution: Harvard ACMG

### Describe the update

Fix a bug that led to a fatal error if no plumes were detected in the inversion region/period. Solves #228. I tested and the out of the box preview runs successfully.


djvaron:
This looks good to me. Thanks James!

Issue #234, opened 2024-07-01
g(P) scaling of errors is squared when it shouldn't be


nicholasbalasus:
### Name and Institution (Required)

Name: Nick Balasus
Institution: Harvard

### Description of your issue or question

g(P) scales the error variance for super observations. I believe it is applied too early here at L154 and should be applied after the squaring of `obs_error` on L182-L183.

https://github.com/geoschem/integrated_methane_inversion/blob/600057807ae827781239418e50c05a5b27645aae/src/inversion_scripts/invert.py#L154

https://github.com/geoschem/integrated_methane_inversion/blob/600057807ae827781239418e50c05a5b27645aae/src/inversion_scripts/invert.py#L182-L183

@laestrada @sarahhancock @megan-he

djvaron:
Good catch, Nick. It looks like this also affects `lognormal_invert.py` through `merge_partial_k.py`:

https://github.com/geoschem/integrated_methane_inversion/blob/600057807ae827781239418e50c05a5b27645aae/src/inversion_scripts/merge_partial_k.py#L89-L94

https://github.com/geoschem/integrated_methane_inversion/blob/600057807ae827781239418e50c05a5b27645aae/src/inversion_scripts/lognormal_invert.py#L85

Since `g(P) < 1`, I think this would artificially depress the observation error when using super-observations, with the same effect as artificially increasing `gamma`.

nicholasbalasus:
Closed by #235, thanks @laestrada!

PR #235, opened 2024-07-02
bugfix for scaling of So

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Applies a fix for a bug where we scale superobservation error standard deviations instead of error variances, as described in #234.

### DIFF
--- a/src/inversion_scripts/invert.py
+++ b/src/inversion_scripts/invert.py
@@ -150,7 +150,10 @@ def do_inversion(
                 for p in obs_GC[:, 4]
             ]
         )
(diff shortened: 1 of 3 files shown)


nicholasbalasus (approved):

djvaron (approved):

nicholasbalasus:
Going to bring this into `dev` so I can bring it into `bugfix/jacobian-perturbation-fix`.

nicholasbalasus:
This fails because `obs_error` is referenced before assignment. Pushing a quick fix to `dev`.

PR #238, opened 2024-07-03
Initial ICI changes

hannahnesser:
Name: Hannah Nesser
Institution: JPL

### Describe the update
This is a set of changes that:
- Allow the IMI to run on different clusters with different schedulers (e.g., PBS vs. SBATCH) or different Python environment tools (e.g., conda vs. pip);
- And, begins to generalize from the IMI to ICI both by changing variable/file/directory names and by beginning to remove hard-coded methane parameters such as methane molar mass and mol/mol -> mixing ratio conversion factors.

The changes include the following (though I may have forgotten something).

PBS/SBATCH flexibility:
- I replaced the config option `UseSlurm` [bool] with `SchedulerType` [str].
- I added a function to common.sh that is run as `submit_job $SchedulerType SaveOutput JobArguments` where `SaveOutput` is a boolean that determines whether or not the output is appended to imi_output.log. It then calls `submit_slurm_job` or `submit_pbs_job` as appropriate. Note: `submit_pbs_job` currently assumes the use of Pleiades and therefore includes a hardcoded "model" keyword that is required by the system.
- I added a function `convert_sbatch_to_pbs` to common.sh that is run in the case that `SchedulerType = "PBS"`. This function searches for all files that have `SBATCH` in their header and then iteratively replaces the `SBATCH` commands with the `PBS` equivalent. This means that any new SBATCH options added to a header should match the formatting in this function or that the sed statement should be expanded to include the new options. 

Some changes that were needed to make these changes work:
- I went through all the SBATCH options and ensured uniform usage (i.e., replaced --time with -t in all cases).
- I removed all `SBATCH -N 1` uses and instead put the node request into `submit_job` since PBS requires that the number of nodes be assigned only once and that it occur along with CPU requests, etc. in the initial job request.

Conda/pip flexibility:
- I replaced CondaEnv and CondaPath with PythonEnv in config.yml. This is now a path to a python environment file that is sourced by the IMI instead of assuming that the environment can be activated by conda. (The conda activation lines are now included in the python environment files.)

Methane/CO2 flexibility:
- I changed many variable names that include TROPOMI to instead include satellite.
- Similarly, I changed many variable names that include methane or CH4 to include either species or nothing.
- I added functions to utils to obtain the mixing ratio conversion factor (1e9 vs 1e6) or the molar mass as a function of species.


hannahnesser:
Hi Sabour,

Great! If you find that something doesn't work, please feel free to fix it. I haven't done a ton of rigorous testing on this because I don't have much of the methane data downloaded (since I'm applying this to CO2).

Issue #237, opened 2024-07-03
misrepresentation of emissions for non-KF inversions longer than one month


nicholasbalasus:
### Name and Institution (Required)

Name: Nick Balasus
Institution: Harvard University

### Description of your issue or question

When a non-kalman-filter inversion runs more than one month, HEMCO emissions files are generated for each month. In places like the preview and visualization notebook, only the first month of emissions are used. So if you ran a 1 year inversion, the emissions reported are just for January.

https://github.com/geoschem/integrated_methane_inversion/blob/d60a8de773c240e26d53ae2d620353ef562e7e57/src/inversion_scripts/imi_preview.py#L424-L426

PR #239, opened 2024-07-04
Bugfix/prior emissions: switch to daily precompiled emissions

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG
 
### Describe the update
This addresses #237. 
When a non-kalman-filter inversion runs more than one month, HEMCO emissions files are generated for each month. In places like the preview and visualization notebook, only the first month of emissions are used. So if you ran a 1 year inversion, the emissions reported are just for January.

This update computes the daily emissions and updates them daily during the jacobian and posterior simulations.

I also add logic to compute the time averaged emissions for the inversion period for python scripts.

### DIFF
--- a/src/inversion_scripts/utils.py
+++ b/src/inversion_scripts/utils.py
@@ -1,13 +1,16 @@
+import os
+import pickle
+from datetime import datetime, timedelta
 import numpy as np
+import xarray as xr
+import cartopy
+import cartopy.crs as ccrs
 from shapely.geometry.polygon import Polygon
 import matplotlib.dates as mdates
 import matplotlib.pyplot as plt
-from pyproj import Geod
-import cartopy
-import cartopy.crs as ccrs
-import pickle
-import csv
 from matplotlib.lines import Line2D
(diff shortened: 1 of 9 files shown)


src/inversion_scripts/utils.py
```diff
@@ -410,3 +413,53 @@ def get_strdate(current_time, date_threshold):
...
+    filtered_files = []
+    for file in filenames:
+        # Extract the date part from the filename
+        date_str = file.split('.')[1]
+        file_date = datetime.strptime(date_str, "%Y%m%d%H%M")
+
+        # Check if the file date is within the specified range
+        if start_date <= file_date <= end_date:
+            # Create a key for the year and month
+            year_month = file_date.strftime("%Y%m")
```
nicholasbalasus, line 435: I think this is leftover.

laestrada: Oops, yup. Cleaned that up and a misleading comment from previous implementation

nicholasbalasus:
Looks good, thanks!

PR #242, opened 2024-07-10
fix postproc_diags to include 0 hour for all tracers and only merge 0th hour

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Postproc_diags.py addresses geoschem not generating the 0th hour of the initial SpeciesConc file for the simulation start date by merging the 0th hour data from either the spinup run or the previous posterior simulation (kalman mode).

This fixes two bugs in postproc_diags.py:

1) Since the addition of multitracer jacobian runs, postproc_diags would only be applied to a single tracer in each jacobian run, rather than all tracers. This caused all other tracers to be assigned NaN values for the 0th hour. Here, we add additional variables to the spinup/posterior SpeciesConc file for each tracer

2) In some cases on a rerun of postproc_diags.py in kalman mode, the following error can happen:
```
xarray.core.merge.MergeError: conflicting values for variable 'SpeciesConcVV_CH4_0001' on objects to be combined. You can skip this check by specifying compat='override'.
```
Really we should not be overriding hours 1-23 in the original SpeciesConc file -- instead we should only overwrite the 0th hour.  Here, we include an update that only overwrites hour 0, instead of all hours of the first simulation day.

### DIFF
--- a/src/inversion_scripts/postproc_diags.py
+++ b/src/inversion_scripts/postproc_diags.py
@@ -1,6 +1,7 @@
 import xarray as xr
 import os
 from joblib import Parallel, delayed
+import numpy as np
 
 def search_file(file_path, search_string):
     """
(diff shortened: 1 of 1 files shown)


megan-he (approved):
Looks good, this worked for global. Thanks Lucas!

PR #243, opened 2024-07-10
Replace sectoral emissions piechart with barchart in viz notebook

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
As the title says, this replaces the pie chart with a bar chart of positive sectoral emissions. Example output shown below:

image


msulprizio (approved):
Looks fantastic!

PR #245, opened 2024-07-10
Clean up overlap between the IMI preview and the prior simulation

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard 

### Describe the update

With the prior (HEMCO standalone) run now enabled in the IMI, we can clean up the IMI code to:

1. Remove the preview run directory. We no longer run a short GEOS-Chem preview simulation to generate prior emissions. These are now obtained from `prior_run/OutputDir`. Therefore, we can remove `preview_run`. We now rename the directory to simply `preview` since it will only contain output PNG and TXT files from the IMI preview.

2. Use the `prior_run` output as input for the emission fields in generating the state vector. We previously used archived files saved out from HEMCO standalone simulations but these were for a fixed date (2019/01/01) and emissions configuration (from GEOS-Chem 14.4.1). Using the output from `prior_run` will allow the state vector to be generated with the actual emissions for the existing IMI setup.

3. Remove `PreviewDryRun` and `RestartFilePreviewPrefix` from config files. These are no longer needed after implementing item 1 above. If users need to do a dry run to download data, they should use the production dry run option before running the prior simulation.

Also included here are minor fixes allowing users to run without specifying `PointSourceDatasets` and `ForcedNativeResolutionElements`


envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
```diff
@@ -182,7 +181,7 @@ AIRS: false
 ##------------------------------------------------------------------
 
 ## Path for IMI runs and output
-OutputPath: "/n/holyscratch01/jacob_lab/$USER"
+OutputPath: "/n/holylfs05/LABS/jacob_lab/$USER"
```
laestrada, line 184: I think we should keep this set to scratch by default to prevent storage bloat of holylfs from folks testing the imi on cannon. Same for the non-global version of this file

src/notebooks/statevector_from_shapefile.ipynb
```diff
@@ -349,8 +348,8 @@
     "gridDir = f'{config[\"Res\"]}_{config[\"RegionID\"]}'\n",
     "land_cover_file = f'{met_token1}.{constYr}0101.CN.{gridRes}.{config[\"RegionID\"]}.nc'\n",
     "land_cover_pth = f'/home/ubuntu/ExtData/GEOS_{gridDir}/{met_token2}/{constYr}/01/{land_cover_file}'\n",
-    "hemco_diag_file = f'HEMCO_sa_diagnostics.{gridRes}.20190101.nc'\n",
-    "hemco_diag_pth = f'/home/ubuntu/ExtData/HEMCO/CH4/v2023-04/HEMCO_SA_Output/{hemco_diag_file}'"
+    "hemco_diag_file = f'HEMCO_sa_diagnostics.{StartDate}0000.nc'\n",
+    "hemco_diag_pth = f'{RunDirs}/prior_run/OutputDir/{hemco_diag_file}'"
```
laestrada, line 352: Does this now require that you run the prior_run before running the notebook? This might be confusing -- how should a user define there region of interest for this, as they dont yet have a custom statevector?

laestrada (commented):
Hi Melissa, this is great -- thanks! Just a couple questions inline.

envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
msulprizio: Good point. Now reverted in https://github.com/geoschem/integrated_methane_inversion/pull/245/commits/3322f5a751611a4e73363a6bd6b9a530bf28d1c3.

src/notebooks/statevector_from_shapefile.ipynb
msulprizio: Yes, it will use the region defined from the lat/lon values in the config file which the user needs to define to run the preview anyway. This just avoids using old, pregenerated HEMCO standalone output for a fixed date.

laestrada (approved):
This looks good to me, thanks Melissa!

PR #250, opened 2024-07-13
Additional changes for global inversion

megan-he:

Name: Megan He
Institution: Harvard ACMG

### Describe the update

This includes a <10% ice cover threshold for generating the global state vector that excludes Greenland, and an option to use a custom periods.csv to define Kalman Filter update periods (e.g. there are an uneven number of days in each month). There are also minor plotting fixes.

### DIFF
--- a/src/components/kalman_component/kalman.sh
+++ b/src/components/kalman_component/kalman.sh
@@ -21,9 +21,13 @@ setup_kf() {
     cp ${InversionPath}/src/notebooks/kf_notebook.ipynb ${RunDirs}/kf_inversions/
     sed -i 's|\/home\/ubuntu\/integrated_methane_inversion\/config.yml|'$ConfigPath'|g' ${RunDirs}/kf_inversions/kf_notebook.ipynb
 
-
     # Define Kalman filter update periods
-    python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
+    if "$MakePeriodsCSV"; then
+        python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
+    else
+        printf "Copying custom periods.csv to the run directory.\n"
+        cp $CustomPeriodsCSV ${RunDirs}/
+    fi
 
     # Create unit scale factor file
     python ${InversionPath}/src/components/kalman_component/make_unit_sf.py $StateVectorFile $RunDirs; wait
--- a/src/components/kalman_component/kalman.sh
+++ b/src/components/kalman_component/kalman.sh
@@ -22,11 +22,11 @@ setup_kf() {
     sed -i 's|\/home\/ubuntu\/integrated_methane_inversion\/config.yml|'$ConfigPath'|g' ${RunDirs}/kf_inversions/kf_notebook.ipynb
 
     # Define Kalman filter update periods
-    if "$MakePeriodsCSV"; then
-        python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
-    else
+    if ! "$MakePeriodsCSV"; then
         printf "Copying custom periods.csv to the run directory.\n"
         cp $CustomPeriodsCSV ${RunDirs}/
+    else
+        python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait 
     fi
 
     # Create unit scale factor file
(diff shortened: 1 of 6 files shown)


src/components/kalman_component/kalman.sh
```diff
@@ -21,9 +21,13 @@ setup_kf() {
...
 
-
     # Define Kalman filter update periods
-    python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
+    if "$MakePeriodsCSV"; then
+        python ${InversionPath}/src/components/kalman_component/make_periods_csv.py $StartDate $EndDate $UpdateFreqDays $RunDirs; wait
+    else
+        printf "Copying custom periods.csv to the run directory.\n"
+        cp $CustomPeriodsCSV ${RunDirs}/
+    fi
```
laestrada, line 30: Nice! I havent thought to much about this, but are there any assumptions built in to the KF assuming a uniform number of days?

Specifically, In the imi preview I think it assumes a certain number of days for calculating superobservation errors

laestrada (changes requested):
Thanks megan! Thoughts are inline!

I think that having custom inversion intervals is a nice feature!

src/inversion_scripts/imi_preview.py
```diff
@@ -310,7 +310,7 @@ def imi_preview(
         lat_bounds=None,
         title="Observation density",
         cbar_label="Number of observations",
-        mask=mask,
+        mask=None,
```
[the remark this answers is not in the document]
megan-he: Fixed in https://github.com/geoschem/integrated_methane_inversion/pull/250/commits/b24cd6daea5071870d1dd75606ddab9514389efc

src/utilities/make_state_vector_file.py
```diff
@@ -167,7 +167,7 @@ def make_state_vector_file(
         hd["lon"] = hd["lon"] - 0.0625
 
     # Select / group fields together
-    lc = (lc["FRLAKE"] + lc["FRLAND"] + lc["FRLANDIC"]).drop("time").squeeze()
+    lc = (lc["FRLAKE"] + lc["FRLAND"] + lc["FRLANDIC"].where(lc["FRLANDIC"] < 0.1,drop=True)).drop("time").squeeze()
```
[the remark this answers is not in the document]
megan-he: Ok, I reverted this change in https://github.com/geoschem/integrated_methane_inversion/pull/250/commits/f07179ec6ef4d847bdb29b140186f68612fb997d

src/components/kalman_component/kalman.sh
megan-he: Yeah you're right! I made some changes in `imi_preview.py` so that the first period in periods.csv is used for the # of days to calculate the number of observations per period and A, but I'm not sure if that's the best way... https://github.com/geoschem/integrated_methane_inversion/pull/250/commits/dfeef5bde96bcbaf5d2109fd49e1420ce788a5e7

src/inversion_scripts/imi_preview.py
```diff
@@ -226,6 +226,10 @@ def imi_preview(
 
     plt.rcParams.update({"font.size": 18})
 
+    # Adjust plot mask if global inversion
+    if not config["isRegional"]:
```
laestrada, line 230: Instead of doing it this way, what do you think of having the logical in-line, since we still use the mask below?

Something like:
```
    plot_field(
        ax,
        prior_kgkm2h,
        cmap=cc.cm.linear_kryw_5_100_c67_r,
        plot_type="pcolormesh",
        vmin=0,
        vmax=14,
        lon_bounds=None,
        lat_bounds=None,
        levels=21,
        title="Prior emissions",
        point_sources=get_point_source_coordinates(config),
        cbar_label="Emissions (kg km$^{-2}$ h$^{-1}$)",
        mask=mask if config["isRegional"] else None,
        only_ROI=False,
    )
```

laestrada (changes requested):
Hi @megan-he,

I think this is almost ready for approval. 

1) I added a suggestion for masking

2) can you add documentation for any of the added config variables

3) For the superobservation days in the preview. I think using the first inversion period number of day is fine (if you wanted you could use the average number of days instead).

4) You need to either make the added config variables in all config files and add them to `sanitize_input_yaml` or make them optional  and assume that make_periods is true (that would be my preference here)

src/inversion_scripts/imi_preview.py
megan-he: Cool, fixed in https://github.com/geoschem/integrated_methane_inversion/pull/250/commits/d96ee0d6a658189cd2f8bb209aca2e3fc077c332

megan-he:
Thanks Lucas, I added more fixes and documentation. I also realize that in the case of a custom periods.csv, then it needs to exist before running the preview and doing those calculations, so I added a fix in https://github.com/geoschem/integrated_methane_inversion/pull/250/commits/bdc892f11ae73c7707084bff7bb88b48f3d44005 (in the default case the calculations use UpdateFreqDays instead).

laestrada (approved):
Looks good, thanks Megan!

PR #261, opened 2024-08-03
bugfix for printing posterior emissions

nicholasbalasus:

Name: Nick Balasus

### Describe the update

* Arguments are passed as a string by default, so the filtering turned `period_df` into an empty DataFrame.
* `posterior_emis_ds` will never have a time dimension, so it can't be selected.

I tested this with this script. It feels like `MakePeriodsCSV` should default to `true` but maybe the IMI isn't set up to handle that based on #256 (unless #260 fixes that).
```
#!/bin/bash

git clone https://github.com/geoschem/integrated_methane_inversion.git
cd integrated_methane_inversion
git checkout 8b5bf2a # bugfix/kf-printing
sed -i -e "s|Test_IMI_Permian|Test_IMI_Permian_KF|g" \
    -e "s|BlendedTROPOMI: false|BlendedTROPOMI: true|g" \
    -e "s|KalmanMode: false|KalmanMode: true|g" \
    -e "s|MakePeriodsCSV: false|MakePeriodsCSV: true|g" \
    -e "s|20180508|20180522|g" \
    -e "s|SetupSpinupRun: false|SetupSpinupRun: true|g" \
    -e "s|SetupJacobianRuns: false|SetupJacobianRuns: true|g" \
    -e "s|SetupInversion: false|SetupInversion: true|g" \
    -e "s|SetupPosteriorRun: false|SetupPosteriorRun: true|g" \
    -e "s|DoSpinup: false|DoSpinup: true|g" \
    -e "s|^DoJacobian: false|DoJacobian: true|g" \
    -e "s|DoInversion: false|DoInversion: true|g" \
    -e "s|DoPosterior: false|DoPosterior: true|g" \
    -e "s|/tropomi|/blended|g" envs/Harvard-Cannon/config.harvard-cannon.yml
sbatch -p huce_cascade -t 2-00:00 run_imi.sh envs/Harvard-Cannon/config.harvard-cannon.yml
```

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -48,7 +48,7 @@ LatMax: 33
 KalmanMode: false
 UpdateFreqDays: 7
 NudgeFactor: 0.1
-MakePeriodsCSV: false
+MakePeriodsCSV: true
 CustomPeriodsCSV: "/path/to/periods.csv"
 
 ## State vector
--- a/envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
+++ b/envs/Harvard-Cannon/config.harvard-cannon.global_inv.yml
@@ -48,7 +48,7 @@ LatMax: 90
 KalmanMode: false
 UpdateFreqDays: 7
 NudgeFactor: 0.1
-MakePeriodsCSV: false
+MakePeriodsCSV: true
 CustomPeriodsCSV: "/path/to/periods.csv"
 
 ## State vector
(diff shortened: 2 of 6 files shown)


laestrada (approved):
@nicholasbalasus Thanks! This looks good.

Also, you are correct `MakePeriodsCsv` should be true in all config files by default. Can you add this to this PR? A bugfix was added in #260 to ensure it is only used when KalmanMode is true

PR #262, opened 2024-08-05
default to End for DiagnFreq

nicholasbalasus:

Name: Nick Balasus
Institution: Harvard

### Describe the update

In non-Kalman filter inversion longer than 32 days (@eastjames @sarahhancock etc.), DiagnFreq was being left as Monthly meaning it wouldn't be overwritten to daily here:
https://github.com/geoschem/integrated_methane_inversion/blob/235ff1443a482b72c4831aab47bfec0abc319820/src/components/prior_component/prior.sh#L75

This means the HEMCO standalone wasn't writing daily files and thus the averaging etc. is probably off by a factor of the number of days in each month.

### DIFF
--- a/src/components/template_component/template.sh
+++ b/src/components/template_component/template.sh
@@ -106,13 +106,8 @@ setup_template() {
         sed -i -e "s:\$RES:\$RES.${RegionID}:g" HEMCO_Config.rc.gmao_metfields
     fi
 
-    # Determine length of inversion period in days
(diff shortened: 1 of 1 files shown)


laestrada (approved):
Thanks Nick!

The time averaging just does:
```
prior_ds = xr.concat(hemco_diags, dim="time")
return prior_ds.mean(dim=["time"])
```
so the averaging would be equally weighting each month in the calculation, so, as you note, slightly off based on the number of days per month

PR #263, opened 2024-08-06
account for GC bounds being grid box edges

nicholasbalasus:

Name: Nick Balasus
Institution: Harvard

### Describe the update

Accounts for GC bounds being grid cell edges. Initially addressed in #190 but fixed fully here.


laestrada (approved):
Looks good to me!

PR #264, opened 2024-08-07
Bugfix/new-env

nicholasbalasus:

Name: Nick Balasus

### Describe the update

With the new conda environment, there are some new warnings in the IMI test case. One is for duplicate dimensions (`nvar`) which is fixed here. 

The other is for NaNs being passed to the `uint8` casting, which is also fixed here. This is with respect to the `surface_classification` variable in TROPOMI files. There are no NaNs in the blended files because QA flag filtering has already occurred.

The posterior was also failing with `Infinity in DO_CLOUD_CONVECTION!` which was due to NaNs in the scale factors. These NaNs are set to 1.

### DIFF
--- a/src/inversion_scripts/operators/TROPOMI_operator.py
+++ b/src/inversion_scripts/operators/TROPOMI_operator.py
@@ -529,7 +529,12 @@ def read_tropomi(filename):
         with xr.open_dataset(filename, group="PRODUCT/SUPPORT_DATA/INPUT_DATA") as tropomi_data:
             dat["methane_profile_apriori"] = tropomi_data["methane_profile_apriori"].values[0, :, :, ::-1]  # mol m-2
             dat["dry_air_subcolumns"] = tropomi_data["dry_air_subcolumns"].values[0, :, :, ::-1]  # mol m-2
-            dat["surface_classification"] = (tropomi_data["surface_classification"].values[0, :, :].astype("uint8") & 0x03).astype(int)
+            sc = tropomi_data["surface_classification"].values[0, :, :]
+            nan_mask = np.isnan(sc)
+            sc_no_nans = np.nan_to_num(sc, nan=0)
+            sc = (sc_no_nans.astype("uint8") & 0x03).astype(int)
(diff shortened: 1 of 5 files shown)


src/inversion_scripts/operators/TROPOMI_operator.py
```diff
@@ -529,7 +529,16 @@ def read_tropomi(filename):
...
-            dat["surface_classification"] = (tropomi_data["surface_classification"].values[0, :, :].astype("uint8") & 0x03).astype(int)
+
+            # Surface classification values of NaN will be filtered out
+            # due to a low QA value. We can make sure by setting them to 5
+            # and making sure nothing outside of [0,1,2,3] makes it through.
+            sc = tropomi_data["surface_classification"].values[0, :, :]
+            nan_mask = np.isnan(sc)
+            sc_no_nans = np.nan_to_num(sc, nan=0)
+            sc = (sc_no_nans.astype("uint8") & 0x03).astype(int)
+            sc[nan_mask] = 5
```
laestrada, line 540: maybe I am missing something, but is there a reason why `nan_mask` exists? Couldn't you do:
```
sc = tropomi_data["surface_classification"].values[0, :, :]
sc_no_nans = np.nan_to_num(sc, nan=5)
sc = (sc_no_nans.astype("uint8") & 0x03).astype(int)
dat["surface_classification"] = sc
```
to the same effect?

laestrada (commented):
Thanks Nick! One comment in line.

src/inversion_scripts/operators/TROPOMI_operator.py
nicholasbalasus: Hi Lucas -- it doesn't matter since these will be filtered out anyway, but this would be cast to 1 which would prevent a potential error check on if your filtered results contain anything that is not [0,1,2,3].

```
import numpy as np
(np.array((5)).astype("uint8") & 0x03).astype(int)
```
> 1

laestrada (approved):
Looks good, thanks!

PR #269, opened 2024-08-27
Fix and improve IMI Preview DOFS calculation

laestrada:

Name: Lucas Estrada and James East
Institution: Harvard ACMG

### Describe the update
IMI preview was assuming the number of superobservations in each state vector element was the number of observation days. This will be an overestimate because some days sv elements will not be observed. The fix was to properly account for the number of superobservations by counting the number of *successful observation days*.

However, once the fix was applied the IMI preview underestimated the actual DOFS. Using a 1 month CONUS inversion as a test case. The original code estimated a DOFS of 19, the updated code found DOFS of 7, and the actual was 29. The systematic underestimate is likely due to only considering emissions having sensitivity to local observations.

To fix this underestimate, we consider observations within the state vector element and observations up to 2 grid cells away from the state vector element. In our test case this estimates a DOFS of 31. This is the approach used in Nesser et al., 2021, except they used observations up to 3 grid cells away.

### DIFF
--- a/src/inversion_scripts/imi_preview.py
+++ b/src/inversion_scripts/imi_preview.py
@@ -572,24 +572,22 @@ def estimate_averaging_kernel(
 
     df_super["lat"] = to_lat(df_super.old_lat)
     df_super["lon"] = to_lon(df_super.old_lon)
-    groups = df_super[['obs_count', 'lat', 'lon']].groupby(['lat', 'lon'])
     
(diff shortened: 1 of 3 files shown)


eastjames (approved):
Thanks, Lucas! I tested your latest changes that use observations in 2 concentric rings for a non-Kalman case and it works. I have no inline comments.

PR #272, opened 2024-09-02
add additional config variables to control inversion step resource allocation

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Optionally allow user to control the resources allocated to the inversion step separately from the `RequestedMemory` and `RequestedCPUs`. If `InversionMemory` and `InversionCPUs` are not set, then the default Memory and CPUs will be used.

### DIFF
--- a/src/components/inversion_component/inversion.sh
+++ b/src/components/inversion_component/inversion.sh
@@ -87,7 +87,7 @@ run_inversion() {
     # Execute inversion driver script
     sbatch --mem $InvMem \
            -c $InvCPU \
-           -t $RequestedTime \
+           -t $InvTime \
            -p $SchedulerPartition \
            -W run_inversion.sh $FirstSimSwitch; wait;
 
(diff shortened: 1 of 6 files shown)


src/components/inversion_component/inversion.sh
```diff
@@ -80,9 +80,13 @@ run_inversion() {
...
+    # Set inversion memory, CPUs, and time
+    InvMem="${InversionMemory:-$RequestedMemory}"
+    InvCPU="${InversionCPUs:-$RequestedCPUs}"
+    InvTime="${InversionTime:-$RequestedTime}"
     # Execute inversion driver script
-    sbatch --mem $RequestedMemory \
-           -c $RequestedCPUs \
+    sbatch --mem $InvMem \
+           -c $InvCPU \
            -t $RequestedTime \
```
megan-he, line 90: Should this be `InvTime`?

megan-he: I think `InversionTime` also needs to be set in the config files. Unless we just use one `RequestedTime` for everything

laestrada: Thanks @megan-he, updated!

PR #274, opened 2024-09-06
Rename 'prior_run' and component scripts to reflect HEMCO standalone simulation and not true prior run

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

There was some confusion about the use of `prior_run` and `prior.sh` which was used for the HEMCO standalone simulation used to generate total prior emissions output for use in the preview, jacobian simulations, etc. The prior run in the inversion typically refers to jacobian run 0000 (i.e. the model run with zero perturbations applied). To avoid confusion, we now rename the `prior_run` directory to `hemco_prior_emis`. The prior component is now referred to as the hemco_prior_emis component.


laestrada (approved):
Looks good! Thanks Melissa!

PR #275, opened 2024-09-06
Bugfix/jacobian sf

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This fixes a bug in the usage of a precomputed jacobian to perform an inversion with an updated prior inventory. This PR addresses that we need to scale the jacobian matrix based on the corresponding perturbation scale factors applied in each inversion.


djvaron (approved):
Looks good, thanks Lucas.

PR #276, opened 2024-09-09
Feature/geoschem14.4.3 update

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Update companion geoschem version for the IMI to version 14.4.3. Importantly, this update to geoschem defaults to using TURBDAY instead of VDIFF for PBL mixing. @nicholasbalasus found VDIFF does not conserve mass. In future versions of geoschem, the default will be switched back to VDIFF once the mass conservation issue is fixed.

### DIFF
--- a/run_imi.sh
+++ b/run_imi.sh
@@ -134,7 +134,7 @@ mkdir -p -v ${RunDirs}
 
 # Set/Collect information about the GEOS-Chem version, IMI version,
 # and TROPOMI processor version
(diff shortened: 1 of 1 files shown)


nicholasbalasus:
Hey Lucas, this change didn't go into 14.4.3. The fix for VDIFF mass conservation is going into 14.5.0. We shouldn't update the version without updating the BCs anyway.

PR #277, opened 2024-09-11
Plot downloads

JohnThomasSEAS:
Name: John

## Plot download support for both normal and KF inversions

This PR incorporates automatic plot downloading via the plot_field function in inversion_scripts/utils.py. The plots download as the viz notebook is run to make the HTML version at the end of processing (i.e. it doesn't require an additional step to download plots, they'll just be there after a successful run).

Two new parameters are added to the plot_field function: 
- save_path: ./plots for normal inversion and ../plots/period_x_plots for KF
- clean_title (optional): some of the title arguments have special scientific characters that didn't play nice in file names so for these, I just removed the special characters

These two arguments are added to every plot_field call in visualization_notebook.ipynb (except in the case of side-by-side plots, in which case they are added to the second in the cell (this just ensures that only one plot is downloaded, not two)).

The added changes won't fundamentally change the viz notebook or the HTML version produced.


laestrada (approved):
Made some minor adjustment, but looks good to merge! Thanks John!

PR #278, opened 2024-09-18
fix boundary condition readme

nicholasbalasus:

Name: Nick Balasus

### Describe the update

Updating outdated README + comments.


nicholasbalasus:
Also removes a test that isn't appropriate when not starting from 1 April 2018.

laestrada (approved):
THanks!

PR #280, opened 2024-10-05
Reduce memory usage in Jacobian creation

eastjames:

Name: James East
Institution: Harvard ACMG

### Describe the update

Small changes to `operator_utilities.py` to decrease its memory usage when creating Jacobian. This should speed up the inversion step. Because it uses less memory, users can use more cores in the inversion step.

Tests on lognormal out-of-the-box Permian test case had identical inversion results. Inversion step was 20% faster and used 40% less memory.


peterivatt:
I'm helping one of our group members with their issue (#314) and noticed some unexpected memory behaviour related to the `xarray.concat` call in `operator_utilities.py`:

```python
ds_sensi = xr.concat(ds_all, "element")
ds_sensi.load()
sensitivities = ds_sensi["ch4"].values
```

Initially, I suspected this would be fixed with this pull request, but removing the .load() step results in only a minimal change in memory usage.

Since there are no large variables in files other that ["ch4"], I would have expected ds_sensi.load() and ds_sensi["ch4"].values to occupy roughly the same memory footprint.

In fact it could even be expected to be more memory without `ds_sensi.load()` as the Dask graph persists, so Dask still retains the task graph in memory.


I'm running a few more tests and will follow up with a more detailed writeup. But it would be great to know if I am missing something here as the improvement seems unexpectedly large and might point to some unintuitive behaviour.

eastjames:
Hi @peterivatt, I also found unintuitive performance when trying to optimize memory usage with xarray and dask in this case. The 40% memory reduction I listed in the PR was for our IMI test case, which is very small. When I implemented this on a much larger production run, performance was worse (I don't remember the specifics), so I changed this PR to a draft.
If you find a more efficient approach, please share! I am not an expert on implementing dask so it could certainly be the case that the current implementation has inefficiencies.

PR #283, opened 2024-10-22
Bugfix/precomputed k sfs

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
The path that the jacobian sfs used to scale K in the event of using a precomputed jacobian was incorrect when doing a standalone inversion.


megan-he (approved):
Looks good, thanks Lucas!

PR #285, opened 2024-10-30
Bugfix for calculating jacobian scale factor

megan-he:

Name: Megan He
Institution: Harvard ACMG

### Describe the update

This fixes the calculation of scale factors to apply to the precomputed jacobian (previously in #275). The scale factor should be calculated as the reference perturbation scale factor divided by the "new" perturbation scale factor (i.e., `ref_pert_sf / pert_sf`. 

This ratio is effectively scaling the jacobian by the new total emissions divided by the reference total emissions: `ref_pert_sf / pert_sf = (target emissions / reference emissions) / (target emissions / new emissions) = new emissions / reference emissions`.


laestrada (approved):
Thanks for catching this Megan -- this makes sense to me!

PR #286, opened 2024-11-08
Additional fixes for using precomputed Jacobian

megan-he:

Name: Megan He
Institution: Harvard ACMG

### Describe the update

Additional fixes when using the precomputed K option include only applying the K scaling factor to emissions elements when doing OH optimization, and defining/making directory paths.

### DIFF
--- a/src/components/jacobian_component/jacobian.sh
+++ b/src/components/jacobian_component/jacobian.sh
@@ -28,13 +28,18 @@ setup_jacobian() {
     # Create directory that will contain all Jacobian run directories
     mkdir -p -v jacobian_runs
 
-    if [ $NumJacobianTracers -gt 1 ]; then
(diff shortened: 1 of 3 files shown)


src/components/jacobian_component/jacobian.sh
```diff
@@ -28,13 +28,18 @@ setup_jacobian() {
...
-    else
-        nRuns=$nElements
+            # Determine approx. number of CH4 tracers per Jacobian run
+            printf "\nCombining Jacobian runs: Generating $nRuns run directories with approx. $NumJacobianTracers CH4 tracers (representing state vector elements) per run\n"
+        else
+            nRuns=$nElements
+        fi
+    else 
+        # only need to set up the prior run directory
+        nRuns=0
```
megan-he, line 42: I think this should affect only the creation of jacobian run dirs in the while loop below when using the PrecomputedJacobian option.

laestrada (approved):
Thanks Megan! This looks good!

PR #290, opened 2024-11-20
fix for misapplication of perturbation delta in jacobian sensitivity calculation

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
When we calculate our jacobian matrix we do the following calculation:
`sensitivity_xch4 = delta_xch4 / perturbation`
where the perturbation is the fractional change from the original emissions eg. 50% perturbation translates to a perturbation = 0.5. 

However, in HEMCO_Config.rc, when we apply the perturbation, a 50% perturbation is represented by a scale factor of 1.5. 

Since the introduction of #226, we did not account for this difference between the representations of the perturbation in HEMCO vs in the jacobian calculation. This introduces an error in the jacobian calculation. Luckily, since the emission perturbations are relatively high (targeting 1e-08 kg/m2/s) the impact on results should be relatively small. For example, if an emission element was perturbed by 52x, the jacobian calculations goes from `sensitivity_xch4 = delta_xch4 / 52` to `sensitivity_xch4 = delta_xch4 / 51`. This expectation is confirmed by my CONUS work for 2019 which goes from 41.075 --> 40.810 Tg/a. Absolute differences shown in the image below.

For OH, we are using a relatively small (10%) perturbation, so this bug matters a lot and would lead to very low sensitivity of OH to the observations in the jacobian. 

This bugfix addresses this issue for both the emission elements and the OH elements.

To avoid recalculating the jacobian if perturbation simulations were already deleted, users can update `invert.py` to do the following rescaling of K:
```
        pert_sfs = np.load("../archive_perturbation_sfs/pert_sf_1.npz")["effective_pert_sf"]
        scaling_matrix = pert_sfs / (pert_sfs - 1)
        
        if bc_optimization:
            # add four elements for BCs
            scaling_matrix = np.append(scaling_matrix, np.ones(4))
        if oh_optimization:
            # add two elements for OH
            scaling_matrix = np.append(scaling_matrix, (1.1 / .1) * np.ones(2))
        
        K *= scaling_matrix
```
image

### DIFF
--- a/src/components/jacobian_component/make_perturbation_sf.py
+++ b/src/components/jacobian_component/make_perturbation_sf.py
@@ -124,6 +124,14 @@ def calculate_sfs(state_vector, emis_prior, target_emission=1e-8, prior_sf=None)
 
     # calculate the effective scale factors
     effective_pert_sf = jacobian_pert_sf / flat_prior_sf
+    
+    # subtract 1.0 from the effective scale factors to get the percent 
+    # change used in the inversion sensitivity calculations
+    # eg. 50% == 0.5
+    # This is is different from the jacobian simulations
+    # where we apply the "jacobian_pert_sf" in HEMCO.
+    # 50% perturbation is represented by SF of 1.5
+    effective_pert_sf = effective_pert_sf - 1.0
 
     # return dictionary of scale factor arrays
     perturbation_dict = {
--- a/src/components/jacobian_component/make_perturbation_sf.py
+++ b/src/components/jacobian_component/make_perturbation_sf.py
@@ -17,7 +17,8 @@ from src.inversion_scripts.utils import get_mean_emissions, get_period_mean_emis
 
 def update_jacobian_perturbation_files(jacobian_dir, state_vector_labels, flat_sf):
     """
-    Update the perturbation files in the jacobian run directories with the new scale factors.
+    Update the perturbation files in the jacobian run directories with the new
+    perturbation scale factors.
     """
     contents = os.listdir(jacobian_dir)
     jacobian_rundirs = [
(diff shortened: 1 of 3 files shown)


src/inversion_scripts/operators/TROPOMI_operator.py
```diff
@@ -257,13 +257,13 @@ def apply_average_tropomi_operator(
...
-                oh_perturbation = config["PerturbValueOH"]
+                oh_perturbation = float(config["PerturbValueOH"]) - 1.0
             else:
-                oh_perturbation = 1.0
+                oh_perturbation = 1.0 # should these be 0.0?
             if config["OptimizeBCs"]:
                 bc_perturbation = config["PerturbValueBCs"]
             else:
-                bc_perturbation = 1.0
+                bc_perturbation = 1.0 # should these be 0.0?
```
laestrada, line 266: @eastjames can you advise on whether these should be 1.0 or 0.0? I imagine 0.0 could give some divide by zero errors.

megan-he: I think this is 1.0 because these are the cases where you're not optimizing OH or BCs so there's no subtracting 1 involved

megan-he (approved):
Looks good to me - thanks Lucas!

src/components/jacobian_component/make_perturbation_sf.py
```diff
@@ -124,6 +124,14 @@ def calculate_sfs(state_vector, emis_prior, target_emission=1e-8, prior_sf=None)
 
     # calculate the effective scale factors
     effective_pert_sf = jacobian_pert_sf / flat_prior_sf
+    
+    # subtract 1.0 from the effective scale factors to get the percent 
+    # change used in the inversion sensitivity calculations
+    # eg. 50% == 0.5
+    # This is is different from the jacobian simulations
+    # where we apply the "jacobian_pert_sf" in HEMCO.
+    # 50% perturbation is represented by SF of 1.5
```
eastjames, line 133: Could you be more precise with the terms "perturbation" and "SF"? It's unclear from this comment how to interpret the code on the next line since the term "sf" or "SF" is used in multiple ways in the code and in the comment

eastjames (changes requested):
I think it would be helpful if the use of "SF" and "perturbation" were more distinct in the variable names and comments in `make_perturbation_sf.py`. But the code itself I approve. I made changes to make the logic more clear for the OH and BC pertubations, so there is never a case where they are set to 1. BC is in ppb so it shouldn't necessarily be set to 1 anyway.

eastjames (approved):
Thanks, Lucas!

PR #293, opened 2024-12-05
Update GEOS-Chem version to 14.6.0 release

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

The GEOS-Chem version is updated to the latest release (14.5.0) in `run_imi.sh`. 

The script `run_boundary_conditions.sh` has also been updated to utilize the latest release version of GCClassic instead of hardcoding to a specific release.


nicholasbalasus:
Hi Melissa – we'll need to create a new set of boundary conditions before merging this because of changes to the methane simulation. Do we have space constraints for hosting multiple versions on AWS @laestrada?

I'd like to switch to a strategy of scaling the GEOS-Chem vertical profiles based on ∆(GC-TROPOMI) as opposed to removing this bias uniformly from each of the vertical layers of GC which could be good to pair with this update.

msulprizio:
Thanks @nicholasbalasus. In that case, would it be better to wait for the 14.5.1 release which will contain a fix for [livestock seasonality](https://github.com/geoschem/geos-chem/issues/2513)?

nicholasbalasus:
I think that makes sense!

msulprizio:
OK. I have changed the title of this PR and removed @nicholasbalasus and @laestrada as reviewers. I'll request reviews again when we're ready to update to 14.5.1 following its release.

msulprizio:
I am closing this PR. See #335 instead which updates the IMI to use 14.6.1.

nicholasbalasus:
Thanks, Melissa! Could you remove the change from the BCs script for now?
I'll address that in a separate PR (presumably for IMI 2.2.1).

msulprizio:
> Thanks, Melissa! Could you remove the change from the BCs script for now? I'll address that in a separate PR (presumably for IMI 2.2.1).

Yep, that script has been restored in commit d9a2b804f9270c74c90b5f29a60eae73f3289b97. Sorry about that!

nicholasbalasus:
Thank you!

PR #297, opened 2024-12-13
Feature/ensemble generation

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG
### Questions for reviewers
- By default, all ensemble members are included in the analysis. However, users can filter out ensemble members from there visualizations by updating an exclusion list in the visualization notebook with the member numbers to exclude. Does anyone have issue with this approach? If so, any alternatives?
- Should any additional visualizations or metrics be included aside from the one shown below?
- Suggestions for updating that default figure in the viz notebook?

### Description of update
This update does the following:
- adds documentation for this feature to the readthedocs site
- vectorizes inversion hyperparameters (`PriorError`, `PriorErrorBCs`, `PriorErrorBufferElements`, `PriorErrorOH`, `ObsError`, and `Gamma`), such that multiple can be input in config.yml and outputted as additional data variables in `inversion_result.nc` and `gridded_posterior.nc`. This works for both lognormal and normal inversions
For example, a user could specify a mix of vector and integer values for hyperparameters:
```
PriorError: [2,  2.5]
PriorErrorBCs: 10.0
PriorErrorBufferElements: 0.5
PriorErrorOH: 0.1
ObsError: [15, 10]
Gamma: [0.1, 1.0]
```
This configuration will create 8 ensemble members. Each data variable for ensemble members will have the suffix `i_ensemble_member`. For example `gridded_posterior_ln.nc` will have the following data variables:
```
 'ScaleFactor'
 'S_post'
 'A'
 'ScaleFactor_1_ensemble_member'
 'ScaleFactor_2_ensemble_member'
 'ScaleFactor_3_ensemble_member'
 'ScaleFactor_4_ensemble_member'
 'ScaleFactor_5_ensemble_member'
 'ScaleFactor_6_ensemble_member'
 'ScaleFactor_7_ensemble_member'
 'ScaleFactor_8_ensemble_member'
 'S_post_1_ensemble_member'
 'S_post_2_ensemble_member'
 'S_post_3_ensemble_member'
 'S_post_4_ensemble_member'
 'S_post_5_ensemble_member'
 'S_post_6_ensemble_member'
 'S_post_7_ensemble_member'
 'S_post_8_ensemble_member'
 'A_1_ensemble_member'
 'A_2_ensemble_member'
 'A_3_ensemble_member'
 'A_4_ensemble_member'
 'A_5_ensemble_member'
 'A_6_ensemble_member'
 'A_7_ensemble_member'
 'A_8_ensemble_member'
```
- Hyperparameters used in the ensemble member are listed in the data variable attributes. For example:
```
>>> ds["ScaleFactor_5_ensemble_member"]
<xarray.DataArray 'ScaleFactor_5_ensemble_member' (lat: 49, lon: 39)> Size: 15kB
[1911 values with dtype=float64]
Coordinates:
  * lon      (lon) float32 156B -110.0 -109.7 -109.4 ... -98.75 -98.44 -98.12
  * lat      (lat) float32 196B 26.0 26.25 26.5 26.75 ... 37.25 37.5 37.75 38.0
Attributes:
    prior_err:     2.0
    obs_err:       15.0
    gamma:         1.0
    prior_err_bc:  10.0
    prior_err_oh:  0.0
    units:         1
```
- Sets the default data variables (that are used in the posterior simulation) to the ensemble member with normalized Ja closest to 1. The default data variables have no suffix, but will have an additional attribute `ensemble_member` specifying which ensemble member they correspond to. For example:
```
ds["ScaleFactor"]
<xarray.DataArray 'ScaleFactor' (lat: 49, lon: 39)> Size: 15kB
[1911 values with dtype=float64]
Coordinates:
  * lon      (lon) float32 156B -110.0 -109.7 -109.4 ... -98.75 -98.44 -98.12
  * lat      (lat) float32 196B 26.0 26.25 26.5 26.75 ... 37.25 37.5 37.75 38.0
Attributes:
    prior_err:        2.0
    obs_err:          10.0
    gamma:            1.0
    prior_err_bc:     10.0
    prior_err_oh:     0.0
    ensemble_member:  6
    units:            1
```
- Fixes a bug in `get_posterior_emissions()` that caused resultant emissions to increase with subsequent calls to this function.
- Creates new summary statistic and visualizations of the ensemble in the viz notebook (includes option to remove ensemble members from the inversion via an exclusion list)
```
Prior         emissions : 0.39176373597940184 Tg/y
Posterior     emissions : 1.3971650306919916 Tg/y (default xhat, Ja/n closest to ~1)
Ensemble mean posterior : 0.8787566295668512  +/- 0.46132713051242463 Tg/yr
```
image
image

I have confirmed that these updates are no diff for both normal and lognormal inversions (except for numerical noise).


msulprizio (approved):
Looks great. Thanks for also updating the documentation for this feature!

src/inversion_scripts/invert.py
```diff
@@ -297,34 +308,162 @@ def do_inversion(
...
+    lat_max = float(sys.argv[8])
+    res = sys.argv[9]
+    jacobian_sf = sys.argv[10]
+
+    # read in config file
+    with open(config_path, "r") as f:
+        config = yaml.safe_load(f)
+
+    # set parameters based on config file
+    is_Regional = ensure_float_list(config["isRegional"])
```
megan-he, line 450: This won't work for a global inversion: if `config["isRegional"]` is `False`, then `is_Regional` = `[0.0]`, and the rest of the script proceeds as if it is a regional inversion. 
Maybe set `is_Regional = config["isRegional"]` so that the boolean gets passed into do_inversion_ensemble?

laestrada: Good catch, fixed!

eastjames:
Thanks @laestrada, this is great and will be super helpful! 2 comments, one you already addressed
1) Use "default" instead of "best" (done)

2)  In my opinion, it’s preferable to have an `ensemble` dimension in the output file and a single variable for ScaleFactor, S_post, A. Then the parameters PriorError, Gamma, etc would also be variables with a single dimension `ensemble`. I think the ability to simply add a new dimension instead of create a new variable is a strength of the netcdf format, and I think it could make the code and file easier to use. With that approach, for example, a variable would look like this:
```
>>> ds["ScaleFactor"]
<xarray.DataArray 'ScaleFactor' (lat: 49, lon: 39, ensemble: 6)> Size: 15kB
[1911 values with dtype=float64]
Coordinates:
  * lon      (lon)      float32 156B -110.0 -109.7 -109.4 ... -98.75 -98.44 -98.12
  * lat      (lat)      float32 196B 26.0 26.25 26.5 26.75 ... 37.25 37.5 37.75 38.0
  * ensemble (ensemble) int     ??B  0 1 2 3 4 5
Attributes:
    units:         1
```
Then one of the parameter vars could be like this
```
>>> ds["obs_error"]
<xarray.DataArray 'obs_error' (ensemble: 6)> Size: 1kB
[6 values with dtype=float64]
Coordinates:
  * ensemble (ensemble) int     ??B  0 1 2 3 4 5
Attributes:
    units:         1
 ```
 
This also means it’s easy to select a single ensemble member and automatically have all the hyperparameters. For example If you run `ds2 = ds.sel(ensemble = 2)` on the dataset in xarray (or analogous command in nco or cdo), then you have selected ensemble member 2 and have all its hyperparameters. So if you want the `obs_error`, it’s `ds2.obs_error`. Scale factors are `ds2.ScaleFactor`, etc, or equivalently `ds.sel(ensemble=2).ScaleFactor`.

For the "default" ensemble member, you only need to know the index number of that ensemble member.

I am volunteering to make these changes if helpful, and I don't think they are major.

What do other folks think? @msulprizio @megan-he @djvaron

laestrada:
@eastjames With this explanation you have convinced me. The only question that I have is whether the posterior simulation would correctly access the default ScaleFactor? How would this work exactly?

eastjames:
@laestrada the ensemble results are stored in files with `_ensemble.nc` appended to their names. The default result is stored in the original files, with no `ensemble` dimension, so the posterior run reads the gridded results file just as before. Updates are pushed to the branch and successfully tested for the normal case, for multiple ensemble members, and for a single inversion with no extra ensemble members.

eastjames (approved):

laestrada:
Now tested the updated approach with both lognormal and normal inversions. Good to merge

Issue #231, opened 2024-06-28
CarbonMapper plume downloads fail in some cases


eastjames:
### Name and Institution (Required)

Name: James East
Institution: Harvard ACMG

### Description of your issue or question

Calls to the Carbon Mapper API fail with a 504 Timeout error in some cases. This is new behavior and I believe it is coming from upstream. I opened an [issue](https://github.com/carbon-mapper/platform-public/issues/6) in the Carbon Mapper issue tracker in case they can help.

eastjames:
Temporarily resolved with https://github.com/geoschem/integrated_methane_inversion/commit/f9a45dddbd902fcf7fd399babbfca0ecbbe2e9b4

PR #302, opened 2025-01-17
Minor fixes to plume data handling code

eastjames:

Name: James East
Institution: Harvard ACMG

### Describe the update

This is a minor update only affecting `src/inversion_scripts/plumes.py`.  It addresses issue #231 and fixes another minor bug. There could be minor effects on the state vector creation if a user used k-means clustering to create the state vector and included IMEO and CarbonMapper plumes in their clustering. Otherwise, no changes.


laestrada (approved):
Thanks! Good to merge!

PR #304, opened 2025-01-23
Add fixes to *invert.py scripts to properly determine number of OH elements

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

In `lognormal_inverty.py` it was automatically assumed that the number of OH elements is 1 when setting `OptimizeOH: true`. This is only the case for regional simulations. For global simulations, OH_element_num should be 2.

In `invert.py`, a fix was needed to select the right element in K when using OH optimization in regional vs global simulations. In addition, the logicals for BC and OH optimization have been renamed in this script for consistency with `lognormal_invert.py`.

Also tagging @megan-he and @laestrada who brought this issue to my attention.

### DIFF
--- a/src/inversion_scripts/invert.py
+++ b/src/inversion_scripts/invert.py
@@ -192,7 +192,7 @@ def do_inversion(
                 scale_factors = np.append(scale_factors, np.ones(4))
             reps = K.shape[0]
             scaling_matrix = np.tile(scale_factors, (reps, 1))
-            if optimize_oh
+            if optimize_oh:
                 if is_Regional:
                     K[:, :-1] *= scaling_matrix
                 else:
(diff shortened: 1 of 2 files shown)


src/inversion_scripts/invert.py
```diff
@@ -186,14 +186,17 @@ def do_inversion(
...
-            if bc_optimization:
+            if optimize_bc:
                 # add (unit) scale factors for BCs
                 # as the last 4 elements of the scaling matrix
                 scale_factors = np.append(scale_factors, np.ones(4))
             reps = K.shape[0]
             scaling_matrix = np.tile(scale_factors, (reps, 1))
-            if oh_optimization:
-                K[:, :-2] *= scaling_matrix
+            if optimize_oh
```
megan-he, line 195: Just missing the colon here

msulprizio: Thanks for catching that! It's now resolved in https://github.com/geoschem/integrated_methane_inversion/pull/304/commits/2c957dc3d140176603f24523238b345668bd18d6.

PR #308, opened 2025-01-24
minor fix to prior sbatch command

mezger6:

Institution: Freelancer, on behalf of 2Celcius

### Describe the update

Added missing output parameter in the prior simulation sbatch command when using pre-computed jacobian

#### Issue addressed

The prior simulation sbatch command ran by the `PreComputedJacobian` branch of the `run_jacobian()` function of `src/components/jacobian_component/jacobian.sh` is expected to output to `imi_output.tmp` , but the `-o` parameter is missing, resulting in an unrecoverable error during the run.


msulprizio (approved):
Thanks also for this fix @mezger6

PR #309, opened 2025-01-28
feature/sectoral_preview_plot

JohnThomasSEAS:
- Adds functionality in the preview to download plot of prior sectoral emissions (bar chart)
- Downloads sectoral emissions csv for use in Integral Earth plot building

Both new output files are exported into the existing preview output dir as preview_sectoral_emissions.png and prior_sectoral_statistics.csv.

Tested on Permian case.

### DIFF
--- a/src/inversion_scripts/imi_preview.py
+++ b/src/inversion_scripts/imi_preview.py
@@ -280,6 +280,8 @@ def imi_preview(
         dpi=150,
     )
 
+   
+
     # Plot albedo
     fig = plt.figure(figsize=(10, 8))
     ax = fig.subplots(1, 1, subplot_kw={"projection": ccrs.PlateCarree()})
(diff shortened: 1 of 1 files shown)


src/inversion_scripts/imi_preview.py
```diff
@@ -449,7 +453,7 @@ def estimate_averaging_kernel(
 
         # use the nudged (prior) emissions for generating averaging kernel estimate
         sf = xr.load_dataset(f"{rundir_path}archive_sf/prior_sf_period{kf_index}.nc")
-        prior_ds = get_mean_emissions(startday, endday, prior_cache)
+        get_mean_emissions(startday, endday, prior_cache)
```
msulprizio, line 456: did you mean to delete `prior_ds =` here?

msulprizio (commented):
I just have one comment. Otherwise looks good!

src/inversion_scripts/imi_preview.py
JohnThomasSEAS: Yeah, I'm not entirely sure why this got changed. Good catch! I'll revert this block to the original and retest my code with a KF config.

msulprizio (approved):

PR #312, opened 2025-02-19
update initial xa vector so that buffer and OH elements are relative …

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This is a bugfix to the lognormal inversion. We initialize the normal elements (buffer elements, BC elements, OH elements) in xn to be ones instead of 0 to start except for the BC elements because those are relative to concentration space rather than scale factors. This is different than in the normal inversion where xa is initialized to 0 to represent the prior. To understand why, it is important to understand  how we are calculating the observation model difference in these inversions.

For both lognormal and normal inversions we calculate the observation - model difference. However, this difference can be calculated in two ways:
1. Using just the prior simulation: y_y_prior_diff - K@xa, where  y_y_prior_diff is the difference between y (tropomi) and the prior simulation. In this scenario, xa is initialized as 0's because the prior is already accounted for in y_y_prior_diff. This is how we do the normal inversion
2. Using a background simulation: y_ybkg_diff-(K@xn), where y_ybkg_diff is the difference between y (tropomi) and the background simulation. Here, the prior simulation is represented when xn is a vector of all 1's because y_ybkg_diff does not take into account the impact of emissions. This is what we do in the lognormal inversion.
 
We use option 2 for the lognormal scenario because we approach the answer iteratively, in each iteration we get a new xn,  which means a new modeled xch4 (ppb). One way to get a new modeled xch4 simulation is to put the new xn through geos-chem and get F(xn), which is painful and takes time. But if we can get y_bkg beforehand, and simply do ybkg+ K@xn, we dont have to wait for geos-chem running in the middle of iterations, since ybkg+ K@xn = F(xn).

The bugfix:
Previously, we erroneously had the buffer elements and OH elements in the initial xn set to 0. This bugfix, initialized them as 1. Notably the BC elements are left to be 0 because they are calculating the delta in concentration instead of scalefactors, so 0 represents sticking to the prior.

### DIFF
--- a/src/inversion_scripts/lognormal_invert.py
+++ b/src/inversion_scripts/lognormal_invert.py
@@ -74,10 +74,10 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         OH_element_num = 0
     BC_element_num = 4 if optimize_bcs else 0
     num_sv_elems = (
-        int(state_vector_labels.max().item()) + OH_element_num + BC_element_num
+        int(state_vector_labels.max().item()) + BC_element_num + OH_element_num
     )
     num_buffer_elems = int(config["nBufferClusters"])
-    num_normal_elems = num_buffer_elems + OH_element_num + BC_element_num
+    num_normal_elems = num_buffer_elems + BC_element_num + OH_element_num
     ds = np.load("full_jacobian_K.npz")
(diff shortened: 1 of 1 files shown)


sarahhancock (approved):
Looks great, Lucas, thanks for implementing this fix so quickly!

PR #315, opened 2025-03-13
only calculate sectoral total if running preview

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Fixes a bug where calling estimate_averaging_kernel would fail if the preview directory was not present.


JohnThomasSEAS (approved):
Tested successfully with test preview of Iran (6 mo)

PR #316, opened 2025-03-17
fix corner case bug where KMeans tries to create a clustering with 0 …

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This fixes a corner case in the clustering algorithm where KMeans would be asked to cluster by assigning 0 labels.


msulprizio (approved):
Good to merge. Thanks!

Issue #314, opened 2025-03-11
Typical memory usage for IMI inversion step


mattloman:
### Name and Institution (Required)

Name: Matt Loman
Institution: University of Rochester

### Description of your issue or question

[env file](https://github.com/user-attachments/files/19198194/bluehive.gcclassic.intel.env.txt)
[config.yml](https://github.com/user-attachments/files/19198195/config.yml.txt)

Hi, I'm running IMI 2.0.0 on local cluster and while the simulation runs ok, I've found that the inversion step takes a surprising amount of memory for each thread (~15GB minimum). If I try to run with less than (15*ncores)GB memory, I get a segfault in jacobian.py ([log file](https://github.com/user-attachments/files/19198270/inversion_segfault.txt)). My advisor @ltmurray mentioned that his IMI runs haven't needed nearly that much, although our settings are different. My 1-week simulation has 1587 state vector elements and 100370 blended TROPOMI observations in the area of interest with expected DOFS around 7.

Is 15GB per thread typical resource usage for IMI inversions?

eastjames:
Hi @mattloman, thanks for opening the issue. We've made some changes to the inversion step in IMI v2.0, specifically by getting rid of a file called `calc_sensi.py`. Depending on when @ltmurray ran IMI, the code base may have changed between his runs and yours. I recommend trying to run the inversion step in serial (not multi-threaded), with a very large amount of memory, depending on what your system has available. If possible, try running with 256gb to 512gb or more, if you have a node available with that amount of local memory. In my experience, it can be slow for very large state vector dimensions.

peterivatt:
Hi @eastjames, the cause of the excessive memory demands appears to be related to `isel` being called after `concat`. Resulting in a `concat` followed by `isel` followed by another `concat` creating a ton of complexity in the Dask graph. The dask profiler also seemed to show this induced a memory leak although I'm not entirely sure why. I posted my results in pull request #318.

PR #318, opened 2025-03-25
Optimized Dask-driven Xarray functions for performance.

peterivatt:
Name: Peter Ivatt
Institution: University of Rochester

In the process of investigating issue #314 and testing the changes proposed in #280 , I implemented a few small changes to `operator_utilities.py` that significantly reduced memory usage.

It appears that as the number of elements increases, memory usage eventually explodes. When running with 1587 elements over the NA grid, the peak memory demand reached 14.3GB, while the underlying data size was only 1.8GB. This can be addressed by moving the hour selection step prior to the first concatenation, which substantially reduces the amount of data that needs to be concatenated. This change reduced memory demands for this run to 3.24GB.

The second improvement involved looping over all datasets before the second concatenation and loading them into memory. Dask performing two concatenations before execution likely created a very complex Dask graph. Since all this data would be loaded into memory anyway, this change allows the second concatenation to be performed by xarray using the more memory efficient NumPy function. As a result, memory usage was further reduced to 2.26GB. As with #280 this would make the .load() calls redundant.

### DIFF
--- a/src/inversion_scripts/operators/operator_utilities.py
+++ b/src/inversion_scripts/operators/operator_utilities.py
@@ -137,8 +137,10 @@ def read_geoschem(date, gc_cache, n_elements, config, build_jacobian=False):
             concat_tracers(k, gc_date, config, v, n_elements)
             for k, v in pert_simulations_dict.items()
         ]
+
+        ds_all = [ds.load() for ds in ds_all]
+
         ds_sensi = xr.concat(ds_all, "element")
-        ds_sensi.load()
 
         sensitivities = ds_sensi["ch4"].values
         # Reshape so the data have dimensions (lon, lat, lev, grid_element)
@@ -149,7 +151,7 @@ def read_geoschem(date, gc_cache, n_elements, config, build_jacobian=False):
         ds_emis_base = concat_tracers(
             "0001", gc_date, config, [0], n_elements, baserun=True
         )
-        ds_emis_base.load()
+
         dat["emis_base_ch4"] = np.einsum("klji->ijlk", ds_emis_base["ch4"].values)
 
         # get OH base, run RunName_0000
@@ -158,7 +160,7 @@ def read_geoschem(date, gc_cache, n_elements, config, build_jacobian=False):
         ds_oh_base = concat_tracers(
             "0000", gc_date, config, [0], n_elements, baserun=True
         )
-        ds_oh_base.load()
+
         dat["oh_base_ch4"] = np.einsum("klji->ijlk", ds_oh_base["ch4"].values)
 
     return dat
@@ -220,13 +222,14 @@ def concat_tracers(run_id, gc_date, config, sv_elems, n_elements, baserun=False)
     if baserun:
         keepvars = ["SpeciesConcVV_CH4"]
 
-    ds_concat = xr.concat([dsmf[v] for v in keepvars], "element").rename("ch4")
-    ds_concat = ds_concat.to_dataset().assign_attrs(dsmf.attrs)
     try:
-        ds_concat = ds_concat.isel(time=gc_date.hour, drop=True)  # subset hour of interest
+        dsmf = dsmf.isel(time=gc_date.hour, drop=True)  # subset hour of interest
     except Exception as e:
         print(f"Run id {run_id}. Failed at {gc_date} with error: {e}", flush=True)
         raise e
+
+    ds_concat = xr.concat([dsmf[v] for v in keepvars], "element").rename("ch4")
+    ds_concat = ds_concat.to_dataset().assign_attrs(dsmf.attrs)
     if not baserun:
         ds_concat = ds_concat.assign_coords({"element": sv_elems})
     return ds_concat


eastjames (approved):
Thank you for this improvement @peterivatt. Tested and everything works as expected.

laestrada:
Thanks @peterivatt and @eastjames!

megan-he:
I tried this update in my global inversion and encountered this error at the `[ds.load() for ds in ds_all]` line, which I suspect may be because when each dataset is opened and read immediately one at a time, that leads to a conflict if multiple processes in the parallelization are trying to access and load data from the same underlying files. I also don't get this error immediately, but it shows up after a few months have been processed. I switched back to the previous version of this code (which opens the nc files lazily and only reads them when needed) and the job finished successfully.

Running this serially also did not raise this error.

```
File "/n/home12/mhe/integrated_methane_inversion/src/inversion_scripts/operators/operator_utilities.py", line 141, in read_geoschem
    ds_all = [ds.load() for ds in ds_all]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/integrated_methane_inversion/src/inversion_scripts/operators/operator_utilities.py", line 141, in <listcomp>
    ds_all = [ds.load() for ds in ds_all]
              ^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/dataset.py", line 863, in load
    evaluated_data: tuple[np.ndarray[Any, Any], ...] = chunkmanager.compute(
                                                       ^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/namedarray/daskmanager.py", line 86, in compute
    return compute(*data, **kwargs)  # type: ignore[no-untyped-call, no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/dask/base.py", line 662, in compute
    results = schedule(dsk, keys, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/indexing.py", line 573, in __array__
    return np.asarray(self.get_duck_array(), dtype=dtype)
                      ^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/indexing.py", line 576, in get_duck_array
    return self.array.get_duck_array()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/indexing.py", line 787, in get_duck_array
    return self.array.get_duck_array()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/indexing.py", line 650, in get_duck_array
    array = self.array[self.key]
            ~~~~~~~~~~^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/backends/netCDF4_.py", line 100, in __getitem__
    return indexing.explicit_indexing_adapter(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/core/indexing.py", line 1014, in explicit_indexing_adapter
    result = raw_indexing_method(raw_key.tuple)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/n/home12/mhe/.conda/envs/imi_env/lib/python3.11/site-packages/xarray/backends/netCDF4_.py", line 113, in _getitem
    array = getitem(original_array, key)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/netCDF4/_netCDF4.pyx", line 4957, in netCDF4._netCDF4.Variable.__getitem__
  File "src/netCDF4/_netCDF4.pyx", line 4560, in netCDF4._netCDF4.Variable.shape.__get__
  File "src/netCDF4/_netCDF4.pyx", line 4512, in netCDF4._netCDF4.Variable._getdims
  File "src/netCDF4/_netCDF4.pyx", line 2142, in netCDF4._netCDF4._inq_vardimid
  File "src/netCDF4/_netCDF4.pyx", line 2134, in netCDF4._netCDF4._inq_varndims
  File "src/netCDF4/_netCDF4.pyx", line 2113, in netCDF4._netCDF4._ensure_nc_success
RuntimeError: NetCDF: Not a valid ID
```

peterivatt:
This is interesting, I wonder if this was the cause of the original memory leak.
When ds.load() is called individually, it crashes if another process accidentally closes the file.
However, when concat is used, Dask may attempt to reopen or recover files, but in doing so, it could cause a memory leak by holding onto failed tasks and references.

You could try changing:
`dsmf = xr.open_dataset("/".join([j_dir, file_stub]), chunks="auto")`
to:
`dsmf = xr.open_dataset("/".join([j_dir, file_stub]), chunks=None)`
inside `concat_tracers`.

This will force xarray to use NumPy as the backend rather than Dask.

I've found this approach to be more stable and in my test case faster (see #319 ).

PR #321, opened 2025-04-03
bugfix/bcs-date

nicholasbalasus:

Name: Nick Balasus

### Describe the update

When assigning TROPOMI data to days for making the smoothed BCs, we use the datetime in the filename. We used the start datetime from the filename, but for the first file, we sometimes need to use the end datetime. It is luck that this hasn't come up before (for example when generating mar2024-may2024 BCs).


laestrada (approved):
Thanks for this Nick!

PR #322, opened 2025-04-09
Turn soil absorption on in background simulation

eastjames:
Name: James East
Institution: Harvard ACMG

### Describe the update

In the lognormal inversion, the background simulation has TROPOMI initial and boundary conditions, OH loss, and no emissions. It should also have soil absorption, but it doesn't. This PR turns on MeMo soil absorption in the background simulation in lognormal inversions. Lognormal inversions done without this fix have an overcorrection, the magnitude is probably small.

Without soil absorption turned on in the background, Kx+background does not match the prior/posterior simulation:
S5_kx_vs_gc_no-soil (2)

With soil absorption in the background simulation, Kx+background matches the prior/posterior simulation:
S5_kx_vs_gc_soil (3)


msulprizio (approved):

laestrada (approved):
Looks good. Thanks James!

PR #324, opened 2025-04-11
Chore/replace awscli commands

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This PR removes usage of the `awscli` for downloading input data. Instead we use `boto3` in python scripts which is much faster. We can almost do away with the aws cli dependency entirely except that geoschem uses it for the dry run data download in [this script](https://github.com/geoschem/geos-chem/blob/main/run/shared/download_data.py). That script could be updated as well, but it would require `boto3` to be installed in geoschem users' python environments. 

One option is to make a IMI specific `geoschem_input_data_download.py` script. @msulprizio thoughts?


msulprizio:
Thanks @laestrada. I'm tagging @yantosca to see if he has any thoughts. Perhaps we could update the GEOS-Chem dry run scripts.

yantosca:
Thanks for tagging me @laestrada and @msulprizio.  A couple of thoughts:

1. We'd need to add boto3 as a dependency to GCPy, which would require a new release on PyPi and conda-forge.
2. The download_data.py script writes a bash script that actually does the data download.  This is so that you can also download from repositories via http/https rather than s3.  Going to boto3 might break that.
3. I probably wouldn't be able to get to this for a while, given other responsibilities.  If either of you would like to try to open a PR for this in the GC repo please go ahead.

msulprizio:
@laestrada I spoke with @yantosca earlier today and we think the best way forward would be to make an IMI-specific download script for now.

laestrada:
That seems like the best option to me as well @msulprizio.

msulprizio:
@laestrada Do you think that script would be best added within the IMI (i.e. in this PR) or in the geos-chem repo?

laestrada:
@msulprizio I think it should be within the IMI repo. I will add it to this PR when I have time.

laestrada:
@msulprizio I have updated the PR to now include a new download script. I tested it on AWS and it works. Can you review?

msulprizio (approved):
Looks great! Thanks for these updates @laestrada.

PR #330, opened 2025-05-03
Use constraint matrix to calculate solution in OH optimization

megan-he:

Name: Megan He
Institution: Harvard ACMG

### Describe the update

When weighting the prior OH error terms in the prior error covariance matrix `Sa` in `invert.py` following Maasakkers et al. (2019), this "constraint matrix" should only be used to calculate the solution xhat (`delta_optimized`). The unweighted errors in the original `Sa` should be used to calculate the posterior error covariance matrix `S_post` and the averaging kernel sensitivity matrix `A`. 

Also addresses OH error weighting in the lognormal inversion code.

### DIFF
--- a/src/inversion_scripts/invert.py
+++ b/src/inversion_scripts/invert.py
@@ -232,6 +232,7 @@ def do_inversion(
     # Inverse of prior error covariance matrix, inv(S_a)
     Sa_diag = np.zeros(n_elements)
     Sa_diag.fill(prior_err**2)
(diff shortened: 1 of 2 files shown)


laestrada (commented):
Thanks Megan -- This looks good! In the lognormal inversion we aren't weighting the OH elements right now. It seems like we should, correct? This could be addressed either in this PR or by someone else in another PR (but we should create an issue).

laestrada (approved):

megan-he:
@laestrada yes, I can edit the PR to address the lognormal inversion as well

Issue #331, opened 2025-05-06
Potential bugs for the calculation for averaging kernel sensitivity and cost estimation in preview?


1Dandan:
### Name and Institution (Required)

Name: Dandan Zhang
Institution: Harvard

### Description of your issue or question

While digesting the source code for [preview](https://github.com/geoschem/integrated_methane_inversion/blob/dev/src/inversion_scripts/imi_preview.py), it looks like there could be bug in:

1) For the function in preview to [estimate averaging kernel sensitivity](https://github.com/geoschem/integrated_methane_inversion/blob/dev/src/inversion_scripts/imi_preview.py#L678), if `L_temp=L_native * size_temp`, and then later emission flux over the ROI with the same state vector label is calculated by https://github.com/geoschem/integrated_methane_inversion/blob/dev/src/inversion_scripts/imi_preview.py#L747, it looks like the emission flux would be `= (total emissions in kg/s) / (size_temp^2 * L_native^2)`, but the emission flux should be `= (total emissions in kg/s) / (size_temp * L_native^2)` as `L_native` is the grid cell side length while `size_temp` is the number of state vector elements? In other words, number of state vector elements of `size_temp` would be divided twice accidentally?

2) For the line at https://github.com/geoschem/integrated_methane_inversion/blob/dev/src/inversion_scripts/imi_preview.py#L187, if the cost scales linearly with the inversion area, it should be `* res_factor^2` instead of `* res_factor` as both `lat` and `lon` are reduced by `res_factor`?

PR #337, opened 2025-06-11
Adapt to use GCHP outputs for IMI

1Dandan:

Name: Dandan Zhang

### Describe the update
These updates are primarily for using stretched GCHP simulation for Jacobian runs in IMI, with several additional features/fixes for both IMI-GCC and IMI-GCHP. All modifications are on top of IMI v2.2.0.

#### Specific modifications for IMI-GCHP Jacobian runs:
As [the feature](https://hemco.readthedocs.io/en/stable/hco-ref-guide/more-examples.html) of scaling grid boxes with specific value from a text file in HEMCO does not work with GCHP, as stated in a ticket of https://github.com/geoschem/GCHP/issues/432, I modified the HEMCO code to apply gridded scaling factor based on the value of state vector ID at https://github.com/geoschem/HEMCO/pull/330. This update of HEMCO would work for both GCC and GCHP. Thus, to minimize the differences between IMI-GCC and IMI-GCHP for synchronizing updates later, I modified the code for creating Jacobian runs in IMI-GCC as well. I tested with a Global GCC 2x2.5 and it works well. The changes would be:

- Preparing the GCHP restart file regridded from a satellite restart file with [a regridding function](https://github.com/1Dandan/integrated_methane_inversion/blob/imi-gchp/src/utilities/common.sh#L205C1-L322C1).
- Creating gridded perturbation scaling factors for emissions by [src/components/jacobian_component/make_perturbation_sf.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-381d5b3fc2fe72618e4efe74a9fc021e4088dc88b15fba1c68ced9b4e327047a) and [gridded OH perturbation](https://github.com/1Dandan/integrated_methane_inversion/blob/imi-gchp/src/utilities/common.sh#L323C1-L392C2) (If OptimizeOH is enabled)
- Selecting a scaling factor value based on state vector ID to apply to that state vector element with the functionality enabled by https://github.com/geoschem/HEMCO/pull/330.
- The `ExtData.rc` and `HEMCO_Config.rc` for GCHP carbon simulation for $\rm CH_4$ should be updated by the PR at https://github.com/geoschem/geos-chem/pull/2893. In that [PR](https://github.com/geoschem/geos-chem/pull/2893), I also fixed some time-range and unit errors I found for GCC carbon-run and methane-run. The updated config would not interfere with other carbon simulation, and thus should be merged generally for carbon fix too.

#### Specific modifications for IMI-GCHP to create a grid file first:

- [GridSpec](https://github.com/LiamBindle/gridspec) and [SparSelt](https://github.com/LiamBindle/sparselt) were used to create state vector file with grid box area calculation in advance as implemented in [src/utilities/make_state_vector_file.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-6c168a846d8278acf284dc4afa78811863b8ca5d59895e02a5a5efc718a2164f).

#### Specific modifications for IMI-GCHP inversion (calculating Jacobian matrix):
- Utilities are updated in [src/inversion_scripts/classify_TROPOMI_obs_to_CSgrids.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-21497d890e95ce6598bdfd8ae3f72957067f7f96a44d8e805f8e793078dae081), [src/inversion_scripts/operators/TROPOMI_operator.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-e434374a5e22e5884514f2a58fe48b33309429f9fa0ad4294a99ad22b208a9f3), and [src/inversion_scripts/operators/operator_utilities.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-e117c14a08cde5462f1ecd9e5429ec95424c6efb26c6aad4c54ffa94a898ee3b).
- Using [SphericalPolygon](https://pypi.org/project/spherical-geometry/) for calculating overlap area for easier handling across the dateline. However, SphericalPolygon is 3D polygon and thus would be much slower than [Shapely Polygon](https://shapely.readthedocs.io/en/stable/reference/shapely.Polygon.html). Currently I am restricting the usage of SphericalPolygon only when the GC grid cross the dateline for GCHP for both `polygon_gchp` and `polygon_tropomi`, while Shapely Polygon was used elsewhere.

#### Fixes/features for both IMI-GCC and IMI-GCHP:

- Use time-averaged outputs for methane concentration and level edge diagnostics instead of instantaneous, as satellite observations are averaged to GC grid hourly to get super observations. Using time-average mode would not be that different or less preferred than the instantaneous mode. Using time-average mode would also eliminate the need to append outputs from spin-up for the first hour diagnostics.
- As a result of eliminating the need of appending outputs from spin-up, I modified the IMI code to disable the output from spinup run as well as we only need a new restart file from spin-up.
- Excluding permanent ice for both original TROPOMI and blended satellite observations using the surface classification from TROPOMI dataset. It is equivalent to Megan's filter as specified in https://github.com/geoschem/integrated_methane_inversion/pull/349 for blended satellite observations, but this filter is applied to both blended and operational TROPOMI dataset in this fix.
- Excluding permanent ice or lakes when constructing state vector file.
- Add an option for optimizing soil when using normal errors. Note this option should not be used with lognormal errors as negative flux is not allowed with lognormal error. This option should not be used for regional IMI with buffer clusters as well, as using median scaling factor could potentially lead to large negative flux is one of the buffer state vector elements is negative.
- `Update_prior_emis.py` at [src/components/posterior_component/update_prior_emis.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-3d37d4a5cd937f20303e162abc9ec5e81795565eda91ccd6ada1449742242c2f) was never used. Instead, this function was called from https://github.com/geoschem/integrated_methane_inversion/blob/main/src/inversion_scripts/utils.py#L552. Thus, I deleted this file.
- The perturbation scaling factor starting at 2000 would interfere with the scaling factors used in CEDS 0.1 for carbon simulations, and thus modify it to start at 3000.
- For `ReDoJacobian`, check if the last-day output file is valid with 24 time entries.
- Format gridded netCDF files to have correct `time` dimension to be compatible with MAPL ExtData to read, including state vector, gridded perturbation scaling factor, gridded OH scaling factor, and prior emission outputs.
- In imi_preview.py, the inversion area cannot be calculated correctly with [the function](https://github.com/geoschem/integrated_methane_inversion/blob/main/src/inversion_scripts/imi_preview.py#L171) used there. Instead, an alternate method of estimating number of grid box is used as implemented in [src/inversion_scripts/imi_preview.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-f40c1bfd264189b9d4ee3f03c2a48806fb96f2e2029f02520e32a755c8ebf9ec). This will fix the error stated in https://github.com/geoschem/integrated_methane_inversion/issues/331 for the cost estimation in `imi_preview.py`.
- Fix the error for calculating averaging kernel density as well as stated in https://github.com/geoschem/integrated_methane_inversion/issues/331.
- The message printed out for TROPOMI file time range would be one day more, and thus confusing. [Fix](https://github.com/1Dandan/integrated_methane_inversion/blob/imi-gchp/src/inversion_scripts/jacobian.py#L121C1-L122C30) it accordingly.
- Add the extraction of `GC_index` during the calculation of Jacobian matrix to be used later for super observation plotting.
- Plotting super observations on GC grid, and add plotting for soil absorption when OptimizeSoil is turned on.

#### Yaml file changes: ([An example yaml file](https://github.com/1Dandan/integrated_methane_inversion/blob/imi-gchp/envs/Harvard-Cannon/config.harvard-cannon.gchp.yml))
- For GCHP specific changes, two additional entries of `UseGCHP` and `STRETCH_GRID` were added to `config.yml`. Need to add two lines of: `UseGCHP: false` and `STRETCH_GRID: false` when using GCC too, as these are mandatory to know which code should be applied. 
- Same applies to `OptimizeSoil` too. We need to add an entry of `OptimizeSoil: false` in `config.yml` if not using `OptimizeSoil`. 
- Use blended TROPOMI dataset by default.
- Need to build a new environment with [a new yaml file](https://github.com/1Dandan/integrated_methane_inversion/blob/imi-gchp/envs/Harvard-Cannon/imi-gchp.yml) supplied to use SphericalPolygon package along with other tools needed for GCHP.

#### Instructions/fixes to use stretched GCHP simulations
Note: Below would be fixed in a new version of GC or MAPL ExtData, and thus you only need to make these modifications on top of the raw GCHP v14.6.2 or GCC v14.6.2.
- After cloning [GCHP v14.6.2](https://github.com/geoschem/GCHP/tree/14.6.2) or [GCC v14.6.2](https://github.com/geoschem/GCClassic/tree/14.6.2) (if you want to use what I modified for GCC) git repository, we need to update the `geos-chem` repository with changes for `HEMCO_Config.rc` and `ExtData.rc` (checking out [this one](https://github.com/1Dandan/geos-chem/tree/imi-gchp-restart)), and update `HEMCO` with changes in [this one](https://github.com/1Dandan/HEMCO/tree/imi-gchp).
- Update the scripts in `<your-gchp-repository>/src/MAPL/gridcomps/ExtData/ExtDataGridCompMod.F90` as (removing the conversion of radians to degrees) (Specific to GCHP): 
```
!call MAPL_ConfigSetAttribute(cflocal,value=temp_real*MAPL_RADIANS_TO_DEGREES, label=trim(COMP_Name)//   MAPL_CF_COMPONENT_SEPARATOR//"TARGET_LON:",rc=status)
call MAPL_ConfigSetAttribute(cflocal,value=temp_real, label=trim(COMP_Name)//                            MAPL_CF_COMPONENT_SEPARATOR//"TARGET_LON:",rc=status)
...
!call MAPL_ConfigSetAttribute(cflocal,value=temp_real*MAPL_RADIANS_TO_DEGREES, label=trim(COMP_Name)//   MAPL_CF_COMPONENT_SEPARATOR//"TARGET_LAT:",rc=status)
call MAPL_ConfigSetAttribute(cflocal,value=temp_real, label=trim(COMP_Name)//                            MAPL_CF_COMPONENT_SEPARATOR//"TARGET_LAT:",rc=status)
```

#### Further features to be added:
- Currently supplying GCHP emission outputs would need an additional day. This would be fixed with a new version of MAPL ExtData with ExtData2G
- The current stretched GCHP simulation itself was problematic for version 14.6.2. Only use the feature for uniform GCHP now.


msulprizio:
@1Dandan Can we close this PR? I assume it's superseded by your other PRs?

PR #338, opened 2025-06-11
Speed up merge_partial_k.py by pre allocating list

eastjames:

Name: James East
Institution: Harvard

### Describe the update

Speed up `merge_partial_k.py` by pre-allocating a list that will contain the jacobian before looping through the files. Previously, we concatenated an array as we went, but this was slow. For a particularly large case for me, `merge_partial_k.py` was taking >60 mins. With this change, it took <3 mins.

Not yet tested with dev.


laestrada (approved):
Nice work! Thanks James!

msulprizio:
I ran the out-of-the-box 7-day Permian case with lognormal errors this update and observed the following speedup:

```
Runtime statistics (s):
              Ref       Dev 
 Setup      : 736       693
 Spinup     : 788       814
 Jacobian   : 1185      637
 Inversion  : 235       235
 Posterior  : 583       423
```

Issue #345, opened 2025-07-10
Finding unique TROPOMI data files fails with duplicates for the period between 8/20/22 and 8/30/22


gchang:
### Name and Institution (Required)

Name: George Chang
Institution: Jet Propulsion Laboratory

### Description of your issue or question

I'm using the IMI version 2.1.0 in docker and running the default Permian Basin run. When I change the time period to 8/20/22 to 8/30/22, it fails an assertion with:

```
[ERROR]   Orbit 25208: 2 files
[ERROR]     S5P_RPRO_L2__CH4____20220825T035306_20220825T053633_25208_03_020500_20230608T101735.nc (Mode: RPRO, Collection: 03, Processor: 020500)
[ERROR]     S5P_RPRO_L2__CH4____20220825T035306_20220825T053633_25208_03_020600_20240722T153153.nc (Mode: RPRO, Collection: 03, Processor: 020600)
[ERROR]   Orbit 25218: 2 files
[ERROR]     S5P_RPRO_L2__CH4____20220825T204800_20220825T223128_25218_03_020500_20230608T160131.nc (Mode: RPRO, Collection: 03, Processor: 020500)
[ERROR]     S5P_RPRO_L2__CH4____20220825T204800_20220825T223128_25218_03_020600_20240724T090323.nc (Mode: RPRO, Collection: 03, Processor: 020600)
[ERROR] len(df) 264, len(df[OrbitNumber].unique()) 142:
[ERROR] Failed to get S3 paths: 
Traceback (most recent call last):
  File "/var/spool/slurmd/job00016/slurm_script", line 224, in get_s3_paths
    assert len(df) == len(df["OrbitNumber"].unique())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
[ERROR] Failed to download TROPOMI data: 
Traceback (most recent call last):
  File "/var/spool/slurmd/job00016/slurm_script", line 256, in download_operational_TROPOMI
    s3_paths = get_s3_paths(start_date, end_date, bucket)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/spool/slurmd/job00016/slurm_script", line 224, in get_s3_paths
    assert len(df) == len(df["OrbitNumber"].unique())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```
[...]

PR #346, opened 2025-07-11
account for multiple processings with different versions

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
As reported in #345, sometimes TROPOMI orbits are processed with two different processor versions. This bugfix ensures that only the file using the latest processor is downloaded.


nicholasbalasus (approved):

nicholasbalasus:
temporary fix since the meeo bucket isn't updated anymore

PR #349, opened 2025-07-18
Filter out permanent snow/ice scenes

megan-he:

Name: Megan He
Institution: Harvard ACMG

### Describe the update

This PR excludes scenes over permanent ice surfaces (e.g. Greenland) by reading in an additional TROPOMI surface classification mask called 0xF9 and filtering out the observations if the `surface_classification == 184` for snow/ice. See [here](https://sentinels.copernicus.eu/documents/247904/2474726/Sentinel-5P-Level-2-Product-User-Manual-Methane.pdf) for reference.


msulprizio:
@megan-he These changes look good to me. I pushed an additional commit to this branch to apply the filter to the standard TROPOMI data as well. Can you have a quick look at it and make sure it looks okay?

msulprizio (approved):

PR #352, opened 2025-08-05
Feature/osse simulation

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This PR implements an Observing System Simulation Experiment (OSSE) mode in the IMI. This allows the user to run an inversion on pseudo TROPOMI observations generated with known "true emissions". This allows users to test under what conditions they can expect to constrain emissions.

Users can generate their "true" emissions by either setting the error to randomly perturb the prior emissions by (according to a normal distribution) or they can provide their own scaling of the prior emissions.

Similarly, users can configure the amount of noise that can be applied to their pseudo observations.

New config variables
.. list-table::
   :widths: 30, 70
   :class: tight-table

   * - ``SimulateObs``
     - Boolean to enable running the IMI with pseudo-observations. Default value is ``false``.
   * - ``DoOSSE``
     - Boolean to run the simulation that pseudo-observations will be generated on. This should be run after the SpinupSimulation. Default value is ``false``.
   * - ``EmisRandomPerturbation``
     - Amount of random perturbation to apply to the prior emissions to generate synthetic observations. Uses a Gaussian distribution to assign, unless ``LognormalErrors`` is set to true, then it uses a log-normal distribution. Default value is ``0.5`` (50%).
   * - ``SimulatedObsError``
     - Amount of random gaussian error to apply to the observations sampled from the OSSE simulation. Default value is ``15`` ppb.
   * - ``CreateAutomaticScaleFactorFile``
      - Boolean to create a scale factor file for the OSSE simulation. This file will be used to define the "true emissions" scaling from the prior emissions. Default value is ``true``.
   * - ``ScaleFactorFile``
      - Path to the scale factor file for the OSSE simulation. This file will be used to define the "true emissions" scaling from the prior emissions. Only used if ``CreateAutomaticScaleFactorFile`` is ``false``.


msulprizio (approved):
This looks great @laestrada. Thanks also for updating the documentation!

PR #355, opened 2025-08-06
To use time-average mode for GC outputs

1Dandan:

Name: Dandan Zhang

### Describe the update
As we are averaging TROPOMI observations into each grid cell over each hour, how about we change the averaging mode of simulation diagnostics to “time-averaged” instead of “instantaneous”? I feel it should be quite similar for hourly outputs between these two modes. In this way, we can avoid filling first missing hour at the start date of simulations. 
GCHP carbon simulation output for instantaneous mode will be offset by one hour for every simulation day, while the time-average mode will always work well. Postprocessing GCHP carbon outputs to have the initial hour could be cumbersome. To minimize the differences between IMI-GCHP and IMI-GCC, how about using time-averaged mode for both?

- Use time-averaged outputs for methane concentration and level edge diagnostics instead of instantaneous. Using time-average mode would also eliminate the need to append outputs from spin-up for the first hour diagnostics.
- As a result of eliminating the need of appending outputs from spin-up, I modified the IMI code to disable the output from spinup run as well as we only need a new restart file from spin-up.
- Add `InvPartition` so that we can add it in the config file for easier request of big-memory node 
- `Update_prior_emis.py` at [src/components/posterior_component/update_prior_emis.py](https://github.com/geoschem/integrated_methane_inversion/pull/337/files#diff-3d37d4a5cd937f20303e162abc9ec5e81795565eda91ccd6ada1449742242c2f) was never used. Instead, this function was called from https://github.com/geoschem/integrated_methane_inversion/blob/main/src/inversion_scripts/utils.py#L552. Thus, I deleted this file.


djvaron:
Using the time-average diagnostics rather than instantaneous makes sense to me.

PR #356, opened 2025-08-06
To use gridded state vector and perturbation scaling factor in HEMCO for IMI-GCC and IMI-GCHP

1Dandan:

Name: Dandan Zhang

### Describe the update

As [the feature](https://hemco.readthedocs.io/en/stable/hco-ref-guide/more-examples.html) of scaling grid boxes with specific value from a text file in HEMCO does not work with GCHP, as stated in a ticket of https://github.com/geoschem/GCHP/issues/432, I modified the HEMCO code to apply gridded scaling factor based on the value of state vector ID at https://github.com/geoschem/HEMCO/pull/330. This update of HEMCO would work for both GCC and GCHP. Thus, to minimize the differences between IMI-GCC and IMI-GCHP for synchronizing updates later, I modified the code for creating Jacobian runs in IMI-GCC as well.
To work with IMI specific configs, the changes made at https://github.com/geoschem/geos-chem/pull/2893 need to be incorporated to make it work properly for creating Jacobian run directories.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -7,6 +7,9 @@ UseSlurm: true
 SafeMode: true
 S3Upload: false
 
+UseGCHP: false
+STRETCH_GRID: false
+
 ## Period of interest
 StartDate: 20180501
 EndDate: 20180508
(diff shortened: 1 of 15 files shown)


lizziel:
Thanks @1Dandan. Could you expand the description to include a bulleted list of all of the changes and why they were made? It looks like changes are showing up beyond the change in perturbation method for Jacobian runs.

PR #357, opened 2025-08-06
Configurations and environment files for IMI-GCHP

1Dandan:

Name: Dandan Zhang

### Describe the update
`config.yml` and environment file for IMI-GCHP

- An example `config.yml` for IMI-GCHP
- Use blended TROPOMI dataset by default.
- Update environment yaml file to use SphericalPolygon package along with other tools needed for GCHP.


msulprizio:
A few comments:

1. Can you remove the `OptimizeSoil` update here? That should come in its own PR alongside the code updates to optimize for soil absorption (and the option should be added to all config files).
2. Can you clarify how the new GCHP environment file (imi-gchp.yml) differs from the existing one (imi_env.yml)? Is there a reason we can't just update the existing imi env file to include the packages needed for GCHP? It would be nice not to have separate environments for IMI with GCHP vs GCClassic.

PR #360, opened 2025-08-08
Add soil optimization option (off by default)

1Dandan:

Name: Dandan Zhang

### Describe the update

Add the option to optimize soil sinks for normal errors.

- Add `OptimizeSoil` option to `config.yml`
It should only be used for normal errors. 
Do not use it when using Lognormal errors as negative flux is not allowed with lognormal errors, 
When using state vector clustering or buffer cluster for regional cases, it could induce applying a large positive scaling factor to a grid cell with negative net flux, and thus a strong negative net flux.
- Using total net flux instead of total emission flux excluding soil sinks
When using optimizing soil option, the perturbation scaling factors for state vector elements would be calculated against total net flux (`EmisCH4_Total`) instead of total emission flux excluding soil sinks (`EmisCH4_Total_ExclSoilAbs`). 
Thus, the addition of `MeMo_SOIL_ABSORPTION` for the cases of prior, background, perturbation run(s) for OH, and posterior run are excluded.
Getting posterior emissions are also updated to apply the scaling factor to soil sinks too when using optimizing soil option.
- Update prior error to be the quadrature of prior error on total emissions and soil sinks
The relative prior error is calculated by:
`sqrt ( total_emission_error ** 2 + soil_sink_error ** 2 ) / (total_net_flux)`
where:
`total_emission_error = total_prior_emissions * relative_prior_error`, 
`soil_sink_error = soil_sinks * relative_prior_error`, 
and `relative_prior_error = 0.5` by default.

### DIFF
--- a/src/components/posterior_component/posterior.sh
+++ b/src/components/posterior_component/posterior.sh
@@ -65,6 +65,7 @@ setup_posterior() {
             -e "/(((MeMo_SOIL_ABSORPTION/i ))).not.UseTotalPriorEmis" \
             -e "/)))MeMo_SOIL_ABSORPTION/a (((.not.UseTotalPriorEmis" \
             HEMCO_Config.rc
+    fi
     # Turn on LevelEdgeDiags output
     # Output daily restarts to avoid trouble at month boundaries
     if "$HourlyCH4"; then


1Dandan:
Hi @laestrada, would you mind adding the update for the cases with state vector clustering?

1Dandan:
Thanks @laestrada. Yeah, also the way we assign the perturbation scaling factor for the buffer state vector elements in regional GCC case. Then I will make this PR ready for review.

src/components/posterior_component/posterior.sh
```diff
@@ -56,14 +56,15 @@ setup_posterior() {
...
-        -e "/(((MeMo_SOIL_ABSORPTION/i ))).not.UseTotalPriorEmis" \
-        -e "/)))MeMo_SOIL_ABSORPTION/a (((.not.UseTotalPriorEmis" \
         -e "s|gridded_posterior.nc|${RunDirs}/inversion/${gridded_posterior_filename}|g" \
         -e "s|GFED                   : on|GFED                   : off|g" HEMCO_Config.rc
-
+    if [ "$OptimizeSoil" != true ]; then
+        sed -i -e "s|EmisCH4_Total|EmisCH4_Total_ExclSoilAbs|g" \
+            -e "/(((MeMo_SOIL_ABSORPTION/i ))).not.UseTotalPriorEmis" \
+            -e "/)))MeMo_SOIL_ABSORPTION/a (((.not.UseTotalPriorEmis" \
+            HEMCO_Config.rc
```
msulprizio, line 67: I believe we need an `fi` here to close the IF block

1Dandan: Yes. Indeed. I add an `fi` now. Thanks Melissa.

Issue #319, opened 2025-03-26
Inefficient nested parallelism


peterivatt:
Name: Peter Ivatt
Institution: University of Rochester

**Issue Explanation:**
The jacobian.py script currently uses Joblib's Parallel to run the TROPOMI operators in parallel, with one joblib worker running on each available core (n_jobs=-1). Each core then calls the `read_geoschem` function, where XArray is used to read and manipulate many NetCDF files. The `chunks` keyword argument is set to `"auto"`, which instructs XArray to use the Dask backend for parallel processing.

This setup leads to a situation where n Joblib workers are each creating their own Dask workers, potentially leading to n^2 Dask workers, which is likely highly inefficient.

**Question:**
When the GC files are individually much smaller than the total available memory, is there any real benefit to using XArray with Dask for parallel file processing when the Numpy backend could be used instead? Setting the `chunks` parameter to `None` would force XArray to use Numpy (without Dask), which has far less overhead.

**Observations:**
In pull request #318, significant memory reductions were achieved by optimizing XArray. When testing the 1.8GB dataset, the peak memory usage was futher reduced from 2.26GB to 1.98GB and ran 30% faster when switching to Numpy as the backend, instead of Dask.

**Discussion:**
Given that the individual GeosChem files aren't particularly large for many users, the benefits of Dask seem minimal. Dask also introduces complexity and can be difficult to profile particualrly when nested inside a joblib worker, making optimization cumbersome.

If Dask is essential for handling very large datasets, perhaps in the config "Dask mode" could be introduced where `n_jobs` is set to a low number, providing many cores per job to effectively manage the large dataset without running into resource congestion.

PR #365, opened 2025-08-28
To reduce massive memory requirement due to the concatenation of all 3D datasets from Jacobian runs

1Dandan:

Name: Dandan Zhang

### Describe the update

The current procedure to calculate Jacobian matrix K is:
first concatenate the 3D datasets from all Jacobian runs as suggested in the [concat_tracers function](https://github.com/geoschem/integrated_methane_inversion/blob/main/src/inversion_scripts/operators/operator_utilities.py#L169), and then calculate the Jacobian matrix for each super observation as stated in the [apply_average_tropomi_operator function](https://github.com/geoschem/integrated_methane_inversion/blob/main/src/inversion_scripts/operators/TROPOMI_operator.py#L226C1-L250C44).
This is very memory-intense to concatenate all the 3D netCDF datasets from Jacobian runs. 
Because of this, currently, we need ~500 GB for the global inversion at 2x2.5 resolution with 47 layers, and thus it becomes almost impossible for an inversion at a finer global resolution.

Considering that we only need 2D virtual TROPOMI column mixing ratio (XCH4), we can avoid the massive memory requirement by first calculating virtual XCH4 for each Jacobian run and then concatenating them to get the virtual XCH4 for all Jacobian runs and hence all state vector elements.

All changes are based on imi-v2.2.0. 
The required changes would be:
1) Remap vertical pressure from GC simulation to TROPOMI vertical layer for all grid cells to obtain vertical remapping weights.
2) Apply that vertical remapping weights for each Jacobian output and then calculate XCH4 for all grid cells.

Additional change is to add a check if a hourly subset of prior simulation is already done or not in `setup_gc_cache.py` to avoid duplicate processing.

### Expected changes
It will reduce the excessive memory requirement from the concatenation of 3D Jacobian outputs, without sacrificing the computing efficiency (comparable wall time with less memory requested).

The memory needed (some test cases):
With changes, for a 1-day global inversion using GCC 2x2.5 with 47 layers and 4368 state vector elements: 101.52 GB (wall time: 00:33:17; requested 300 GB, 32 cores)
By default, for a 1-day global inversion using GCC 2x2.5 with 47 layers and 4368 state vector elements: 502.48 GB (wall time: 00:36:17; requested 1000 GB, 32 cores)
With changes, for a 1-day global inversion using GCHP C90 stretching factor 4 with 72 layers and 6733 state vector elements: 214.56 GB (wall time: 06:31:53; requested 300 GB, 32 cores)
By default, for a 1-day global inversion using GCHP C48 with 72 layers and 4743 state vector elements: 698.21 GB (wall time: 00:21:48; requested 1000 GB, 32 cores)

The results should be the same.
To validate the changes, I used a 1-day global inversion with GCC 2x2.5 with 47 layers and 4368 state vector elements as a test case. The visualization notebook produced showed that they are essentially the same (with rounding errors). They can be accessed here:
[visualization_notebook_CalK.zip](https://github.com/user-attachments/files/22049838/visualization_notebook_CalK.zip)


msulprizio:
@eastjames Did you still want to review this?

msulprizio:
@1Dandan Would you be able to merge these updates into the latest `dev` branch? I tried doing so, but there are several conflicts in TROPOMI_operator.py, mainly due to the OSSE option that was implemented since your updates, and I'm not sure how to resolve them.

eastjames (changes requested):
Thanks, @1Dandan for solving a problem we've had for a while! I think this will also address #314 and #319. Since this PR would cause `concat_tracers` to no longer be used, and `apply_tropomi_operator` is only used for visualization and therefore only called with `build_jacobian = False`, could this PR also:

- get rid of `concat_tracers` function entirely, since it is no longer used
- get rid of `if build_jacobian` block inside `read_geoschem` since it is also no longer used

Or that may be better as a separate pull request.

1Dandan:
Thanks @eastjames for reviewing it! 
Would we need `concat_tracers` function for other non-TROPOMI satellite?
For disabling `build_jacobian`, sure, I will delete this option for `apply_tropomi_operator` as we never use this option to be true.

eastjames:
Hi @1Dandan, we currently don't need it for non-tropomi satellites, and if we added satellites in the future I think we would avoid using that function.

1Dandan:
Hi @msulprizio, I have updated the code to adapt for OSSE option.
@laestrada, would you mind reviewing the part for the calculation for OSSE option?

1Dandan:
Thanks @eastjames. I just deleted `concat_tracers` function as suggested.

laestrada:
OSSE updates look OK to me -- have you tested it?

1Dandan:
Thanks @laestrada for reviewing. I tested the original code without OSSE option, but I did not test the OSSE-specific code.

eastjames (approved):

PR #367, opened 2025-09-24
Fix state vector file generation to only include offshore grid boxes when oil/gas emissions are non-zero

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

In PR #348, we switched from using oil+gas emissions to total emissions when generating the state vector netCDF file. This inadvertently includes more ocean grid boxes than needed because some emission sectors include non-zero emissions over oceans simply because they have been regridded from coarse resolution (e.g. 4x5 for termites) datasets. To resolve this, the code has been updated to only include offshore emissions grid boxes where oil+gas emissions are non-zero.

This PR also includes other minor fixes and error checks:

- Add checks to make sure template run directories and HEMCO standalone run directory exist when needed
- Add check to make sure state vector file exists when the create state vector option is turned off
- Update HEMCO prior emissions run to save out single diagnostic file containing average emissions for the inversion period, rather than daily files
- Update IMI run script to send the user an email when the entire IMI job has completed (when submitting via SLURM)

Example StateVector.nc file over Algeria before the fix:

Screenshot 2025-09-24 at 10 39 09 AM

after the fix:

Screenshot 2025-09-24 at 10 36 53 AM

### DIFF
--- a/run_imi.sh
+++ b/run_imi.sh
@@ -3,6 +3,7 @@
 #SBATCH -N 1
 #SBATCH -c 1
 #SBATCH --mem=2000
+#SBATCH --mail-type=END
 #SBATCH -o "imi_output.log"
 
 # This script will run the Integrated Methane Inversion (IMI) with GEOS-Chem.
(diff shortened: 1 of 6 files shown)


laestrada (approved):
Looks good -- thanks!

PR #368, opened 2025-09-30
Revert HEMCO prior emissions output to `Daily` instead of `End`

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

In https://github.com/geoschem/integrated_methane_inversion/pull/367/files, I had changed the HEMCO prior emissions run to output a single file at the end of the run. It turns out we should be saving out daily files so that the jacobian simulations can input total daily emissions.


msulprizio:
Tagging @1Dandan for review

1Dandan:
Thanks @msulprizio. It looks good to me.

PR #371, opened 2025-10-24
Minor math corrections to lognormal_invert

mattloman:

Name: Matt Loman
Institution: University of Rochester

### Describe the update

I've made a couple edits to lognormal_invert.py that I think are necessary to get the math working properly.
- The `prior_scale` variable is $e^{\frac{-s'}{2}}$, the scaling factor applied to $x_{mean}$ to convert to $x_{median}$ in equation (5) of [Hancock et al., 2025](doi.org/10.5194/acp-25-797-2025). Since $K=\partial y/\partial x$, `K_ROI` should be divided by `prior_scale`, rather than multiplied, so that it represents $\partial y/\partial x_{median}$.
- In order for `(y_ybkg_diff - K_full @ xn)` on line 263 to represent the difference between the observations and (when `xn=xa`) the prior run, the lognormal part of `xn` needs to be a median to agree with $K_{ROI}=\partial y/\partial x_{median}$ above, so `xa` needs `prior_scale` applied.
- Finally, since we're no longer treating both $x_{a,mean}$ and $x_{a,median}$ as vectors of ones, the calculation of `xnmean` needs to be updated based on equation (6) of [Hancock et al., 2025](https://acp.copernicus.org/articles/25/797/2025/).

I haven't had a chance to run simulations to compare results, so please let me know if that will be necessary before merging.

### DIFF
--- a/src/inversion_scripts/lognormal_invert.py
+++ b/src/inversion_scripts/lognormal_invert.py
@@ -151,7 +151,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         # the lognormal part of K gets scaled by prior_scale to convert to median
         K_ROI = K_temp[:, :-num_normal_elems]
         K_normal = K_temp[:, -num_normal_elems:]
-        K_ROI = prior_scale * K_ROI
+        K_ROI = K_ROI / prior_scale
         K_full = np.concatenate((K_ROI, K_normal), axis=1)
 
         m, n = np.shape(K_ROI)
@@ -159,7 +159,8 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         # Create base xa and lnxa matrices
         # Note: the resulting xa vector has lognormal elements until the
         # final Buffer, BCs, and OH elements
-        xa = np.ones((n, 1)) * 1.0
+        # Convert to median before calculating lnxa
+        xa = np.ones((n, 1)) * prior_scale
         lnxa = np.log(xa)
 
         # Create normal elements for buffer, BCs, and OH
@@ -295,7 +296,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         dlns = np.diag(lns[:-num_normal_elems, :-num_normal_elems])
         xnmean = np.concatenate(
             (
-                xn[:-num_normal_elems]
+                ( xn[:-num_normal_elems] / xa[:-num_normal_elems] )
                 * np.expand_dims(np.exp(dlns * (0.5)) * prior_scale, axis=1),
                 xn[-num_normal_elems:],
             )


msulprizio:
Thanks @mattloman! I'm tagging @laestrada and @sarahhancock so they can review before we merge this in.

sarahhancock:
Hi @mattloman, thank you for thinking closely about this! I'm open to being convinced on this because the right way to do the mean/median conversion was a big debate in our group this summer. @xlwangg may have thoughts as she looked closely at this too. But here is our logic for the original setup:

To the first point about scaling K, because of how we define K (change in concentrations / relative change in X), scaling K should be done as K_new = K * (new prior / old prior). This makes sense intuitively because, for example, if our new prior estimate has higher emissions for a particular state vector element than our old prior, K should be higher because the relative perturbation in emissions would be the same, but the magnitude of the emissions would be higher so we get a higher change in concentrations.
In this case, new prior = old_prior*prior_scale
So we do K_new = K* (old_prior*prior_scale/old_prior) = K*prior_scale

To your second point, we don't apply prior_scale to xa later because of how we define K. K is already scaled here and is defined as (change in concentrations/relative change in median prior), so K*xa becomes the equivalent of the prior run with the median emissions when xa is a vector of ones.

mattloman:
Hi @sarahhancock, thanks for the detailed explanation. This was something that had been confusing me as I was reading through the code, and your responses help me make sense of it as it's currently written. I'm also less than 100% certain on how this should be addressed, but having read through your responses I've still got a couple thoughts / concerns:

- Simplest first: I do still think that there needs to be a change to the calculation of `xnmean`, since the equation in your paper contains both $x_{a,mean}$ and $x_{a,median}$. If `prior_scale` isn't 1, then one of the two should be in there.

After some consideration, I think the issue I was trying to fix in lognormal_invert.py may still be cropping up in my commit as well. In the current code:
- Defining K as (change in concentrations)/(relative change in median emissions) works fine for `xa`, but once you converge on an optimal solution, the vector of mean scaling factors to minimize $J(x)$ would be `xn / prior_scale` rather than `xnmean` as it's currently calculated. In other words, if mean scaling factors for emissions in each grid cell are the thing we're trying to optimize, I don't think that formation of K allows us to properly close the gap of `(y_ybkg_diff - K_full @ xn)` since `K_full @ xn` isn't actually equal to $F(x_n)$ like it needs to be per Rodgers 2000.

In my commit:
  - I was treating K as (change in concentrations from a simulation with mean emissions) / (relative change in median emissions corresponding to that mean) and converting `xa` to a median in an attempt to resolve that issue, but I think it runs into a similar problem. Although it works fine for `xa`, the conversion from mean to median should be different for all other `xn`, so `K_full @ xn` still wouldn't be equal to $F(x_n)$.
 
My immediate reaction is that we may need to be calculating either `xnmean` or something like `xn_scale` similar to `prior_scale` every iteration to resolve this, but I'd like to hear what others are thinking, especially since it sounds like you all may have spent more time on this than I have.

sarahhancock:
Hi @mattloman, sorry for the slow response and also for the weird formatting of my original reply - I'm not sure how that happened and I'm glad you figured out what I was trying to say!

I think we might be in agreement but just saying things in a different way, but again, I will also let others weigh in on this and I'm also happy to discuss further. The lognormal algorithm optimizes the median emissions, but we don't think our prior emissions represent the median of the lognormal pdf, they represent the mean. So before doing the lognormal algorithm, we convert to the median (and we have to do this mean-> median conversion to K because of how we define it in relative terms). Then we do the lognormal algorithm with everything in lognormal / median space. We are optimizing xn_median - this might be where the disconnect is, you say:
> * if mean scaling factors for emissions in each grid cell are the thing we're trying to optimize

but we have to optimize the median scale factors because of the nature of the lognormal algorithm.
The problem is that in reality, we don't want to report the median of emissions, so we have to convert our optimized median back to the mean.

I also found that in my work, this mean/median conversion had very little impact on my results. I'm curious if it is causing issues for you or if something else might be because again - there are lots of points of confusion in this lognormal algorithm and there could be something else we are missing.

Let me know if this doesn't make sense or if I am totally misunderstanding what you are saying!

mattloman:
Hi @sarahhancock, no worries, thanks for the clarification and interested to hear what the others have to say as well. I think that was my main misunderstanding when I first started looking at lognormal_invert.py, it wasn't clear to me why it was only K getting converted or why it was being done in that way.

Sorry for the very poor wording on my part with optimizing mean vs median - yes, we're optimizing the median, but like you say, we want to do that in a way that we can still back out the mean at the end and have that mean correctly represent the optimized emissions based on the observations we have.

I haven't seen significant impacts from the conversion either, in part because I've been using a pretty small value for $\sigma_g$ (1.3). I started looking at lognormal_invert.py to try to find the solution to what ended up being a different problem, which we've since solved, but I wanted to communicate to you all the concerns I had with the methodology.

I'm not sure I've done a good job so far explaining why I think the math may be off, so let me give it another shot:

K is initially calculated as (change in simulated observations)/(relative change in mean emissions). To convert to median, the assumption is that (change in simulated observations)/(relative change in median emissions) is equivalent to (change in simulated observations)/[relative change in (mean emissions * `prior_scale`)] like it is for xa. If this assumption holds, that addresses most of my concerns (except the calculation of `xnmean`).

What I'm having trouble with is that it seems like this assumption is no longer accurate once xn has been updated to something other than xa, since the conversion from mean to median is dependent on variance which also changes during iteration. This is relevant because median emissions are not directly related to simulated observations, but through their relationship with the mean emissions. The relationship between simulated observations and mean emissions as described by the original K stays constant throughout the inversion, so K doesn't have to be recalculated. But the relationship between the median emissions and mean emissions changes with iteration. This is why I was saying above that `K_full @ xn` wasn't correctly representing $F(x_n)$.

As I mentioned in my previous comment, I think I ended up recreating the same issue from a different angle with the edits on this commit.

msulprizio:
Hi @mattloman. What is the status of this pull request? Should it still be merged into the standard code or should we close it?

PR #372, opened 2025-10-27
Update mapping of default ensemble member based on filtering

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
The IMI filters ensemble members based on the calculated normalized Ja value. However, the default member specified in the inversion_result attributes was not getting updated after filtering. Here we update the index.


msulprizio (approved):
Looks good! Feel free to merge.

PR #373, opened 2025-10-31
Update IMI config to accept additional scheduler options

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This update changes the `UseSlurm` config option to `Scheduler`. It also adds if statements to job submission locations, such that the IMI will run commands interactively if the slurm scheduler isnt available. This is an initial step in generalizing the IMI config file for usage on other clusters.

Note that I have not tested the changes implemented here.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -3,7 +3,7 @@
 
 ## General
 RunName: "Test_Permian_1week"
-UseSlurm: true
+Scheduler: "slurm"
 SafeMode: true
 S3Upload: false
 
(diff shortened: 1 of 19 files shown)


msulprizio:
Thanks @laestrada. These changes look good, but note that this will cause conflicts with similar changes made in https://github.com/geoschem/integrated_methane_inversion/pull/238/files by @hannahnesser. There she uses `SchedulerType` instead of `Scheduler`.

laestrada:
Deleting this branch as it is covered by changes in #238

PR #374, opened 2025-10-31
improve sanitation script to check type and value of config variables

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This PR updates the IMI `sanitize_input_yaml.py` file to perform checks on the validity of the config file based on expected types and values. This is in addition to previously checking for missing config variables.


msulprizio (approved):
Really useful update! Thanks!

PR #378, opened 2025-11-21
Add fixes to netCDF output files and visualization notebook for ensemble mode

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

The code in `lognormal_invert.py` has been fixed to now save out the mean of ensemble members to the `inversion_result_ln.nc` file. This is now consistent with updates made in https://github.com/geoschem/integrated_methane_inversion/pull/361. The code in `lognormal_invert.py` and `invert.py` has also been updated to specify the `long_name` attribute for fields saved out to the netCDF files. Documentation for the ensemble option has been updated to clarify that the ensemble mean is used for the posterior simulation.

Additionally, the visualization notebook has been updated to sum and plot the `EmisCH4_Total_ExclSoilAbs` when comparing the ensemble members and the ensemble mean vs the prior. Previously, `EmisCH4_Total` was used which includes the soil sink flux and therefore made the total emissions value appear lower than it actually is.

### DIFF
--- a/docs/source/advanced/inversion-ensemble.rst
+++ b/docs/source/advanced/inversion-ensemble.rst
@@ -7,40 +7,38 @@ can easily be reused for additional  inversions. This can be useful for understa
 sensitivity of the inversion results to the choice of prior error, observation error, 
 or other inversion parameters.
 
-The IMI has multiple options for creating an inversion ensemble.
-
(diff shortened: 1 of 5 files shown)


laestrada (approved):
Looks good, thanks!

PR #379, opened 2025-12-04
To include the chemical mechanisms for Jacobian tracers with carbon mechanism

1Dandan:

Name: Dandan Zhang

### Describe the update

The chemical mechanism using carbon mechanism is handled by the KPP equations.
The geos-chem changes needed for the Jacobian tracer KPP equations are included in https://github.com/geoschem/geos-chem/pull/3080.
Here it to adapt IMI compilation code to accommodate changes needed:
- scripts to create a series of KPP equations for any number of Jacobian tracers (`src/utilities/expand_carbon_eqn.py`). 
- Now the GEOS-Chem exec is built in a new build folder within the source code of `GCClassic`, and `GEOSChem_build` folder is no longer supported. All GEOS-Chem exec was changed for that linking path.
- As the Jacobian tracers are numbered locally for each perturbation run due to the changes in https://github.com/geoschem/geos-chem/pull/3080, the naming for the added species name is also changed to reflect this in `src/components/jacobian_component/jacobian.sh` and `src/inversion_scripts/operators/operator_utilities.py`
- Another minor change is to update the print_stats in `src/utilities/common.sh` to be robust when certain component is set to false and thus the end time for that component may not be defined.


msulprizio (approved):

msulprizio (changes requested):
The IMI does not currently work with these updates. It crashes in the inversion stage (specifically `jacobian.py`) at these lines in `satellite_operator.py`:

```
    # Construct the list of species vars to request
    keepvars = [f"SpeciesConcVV_{config['Species']}_{i:04}" for i in sv_elems]
```

I believe that needs to be changed to:
```
    keepvars = [f"SpeciesConcVV_{config['Species']}_jac{k:04}" for i in sv_elems]
```

but we also need to add the correct indexing for the jacobian species there. 

@1Dandan  or @lizziel Can you rebase this atop the latest updates and `dev` and test for GCClassic and GCHP?

For now, I will not merge this PR. Instead, we'll keep the number of jacobian tracers per run to 1 for GCClassic so that we can release 2.3.0. This update, when ready, can come into 3.2.1 or later.

msulprizio:
Following @1Dandan's suggestion to add the following lines to `satellite_operator.py` gets past the inversion error.

```
# Local tracer indices are 1..len(sv_elems) within this run
local_indices = range(1, len(sv_elems) + 1)
keepvars = [f"SpeciesConcVV_CH4_jac{idx:04d}" for idx in local_indices]
```

However, now the inversion results are altered and possibly suggestive of an indexing issue. In the out-of-the-box ` 1-week Permian inversion, the scale factors in Ref and Dev are:

scale_factors_ref
scale_factors_dev

I have removed the 2.3.0 milestone for now until we sort this out.

PR #392, opened 2026-02-20
Introduce CH4 and CO2 hierarchical structures into the IMI config file

laestrada:

Name: Lucas Estrada 
Institution: Harvard ACMG

### Describe the update

This PR introduces `CH4` and `CO2` hierarchical structures into the IMI config file by standardizing config handling across IMI with a single config loader and routing runtime scripts through it. This is an incremental update to allow for more specific CO2 configuration options during ICI development without breaking existing CH4 functionality.

- Added `CH4:` and `CO2:` hierarchical config sections for species-specific settings.
- Updated templates/docs to reflect the new structure.
- Added normalization so runtime consumers still receive expected flat keys after loading the active species section.

Example of new CO2 and CH4 hierarchy:
```
CH4:
  SatelliteProduct: "BlendedTROPOMI"
  ## Use observations over water? Set to false to filter out water observations
  UseWaterObs: false
  OptimizeOH: false
  # optimize soil option when using normal errors,
  # do not use it when using Lognormal errors as negative flux is not allowed with lognormal errors,
  # current version is not compatible with state vector clustering
  # either through clustering algorithm or buffer cluster for regional case
  OptimizeSoil: false
  PriorErrorOH: [0.1]
  # Save out GEOS-Chem diagnostics for additional observation operators
  AdditionalDiagnostics: ["ObsPack"]  ## GOSAT, TCCON, AIRS, PLANEFLIGHT

CO2:
  SatelliteProduct: "OCO2"
  UseWaterObs: false
  # Save out GEOS-Chem diagnostics for additional observation operators
  AdditionalDiagnostics: ["ObsPack"]  ## TCCON, PLANEFLIGHT
``` 

Note: I ran an IMI preview, and it ran successfully, but that is the extent of testing done so far.


msulprizio (approved):
Thanks @laestrada! These changes look good. I'll run them through an IMI benchmark run to ensure nothing is broken before we merge into `dev`.

msulprizio:
@laestrada I encountered an error in visualization notebook during the posterior phase of the IMI. I'll push a fix then I think we should be good to merge.

msulprizio:
@laestrada I just pushed a fix for the visualization notebook so this PR now passes a benchmark run. We should be good to merge now.

PR #402, opened 2026-04-24
Turn on soil absorption optimization by default

msulprizio:

Name: Melissa Sulprizio
Institution: Harvard

### Describe the update

The recommendation from the IMI Steering Committee is to have soil optimization turned on by default. It was previously set to `false`. All config files have been updated to set `SoilOptimization: true`. 

Note that users should disable this option if `LognormalErrors: true` because negative fluxes are not allowed with lognormal errors. An error check has been added to `run_imi.sh` after reading the config.yml file to make sure `SoilOptimizarion` and `LognormalErrors` are not true at the same time.


msulprizio:
Turning on soil absorption optimization is causing a handful of jacobian runs to fail with negative concentrations in the perturbed CH4 tracer. From one such GEOS-Chem log file:

```
---> DATE: 2018/05/07  UTC: 04:20
---> DATE: 2018/05/07  UTC: 04:25
 WARNING: Negative concentration for species CH4_0068 at (I,J,L) =            7          35           2
 WARNING: Negative concentration for species CH4_0068 at (I,J,L) =            7          35           3
 WARNING: Negative concentration for species CH4_0068 at (I,J,L) =            7          35           4
 WARNING: Negative concentration for species CH4_0068 at (I,J,L) =            7          35           5
 ...
 ===============================================================================
GEOS-Chem ERROR:
 -> at DO_TEND (in module GeosCore/mixing_mod.F90)
===============================================================================

===============================================================================
GEOS-Chem ERROR: Error encountred in "DO_TEND"!
 -> at DO_MIXING (in module GeosCore/mixing_mod.F90)
===============================================================================

===============================================================================
GEOS-CHEM ERROR: Error encountered in "Do_Mixing"!
STOP at  -> at GEOS-Chem (in GeosCore/main.F90)
===============================================================================
     - CLEANUP: deallocating arrays now...
 ```

I will investigate this further but don't want it to hold up the release of IMI 2.3.0.

msulprizio:
Following discussion among Harvard ACMG members, it has been decided that:

>For methane it seems better to optimize a scale factor. But a problem we ran into when optimizing the soil sink is zero crossing within the cluster, resulting in negative concentrations. We are now solving this problem by not including the soil sink in the perturbation (so that the perturbation stays positive) and diagnosing the soil sink in post-processing like in [He et al. (2026)](https://www.science.org/doi/10.1126/sciadv.adz9007).

PR #411, opened 2026-05-14
Miscellaneous bugfixes for Kalman filter inversion

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
This PR updates:
- directory names that were incorrect. This was missed because they are only used during the Kalman filter inversion. 
- The perturbations netcdf file was incorrectly being applied in the Kalman filter. This file is regenerated on each period of the Kalman filter and saved into `archive_perturbation_sfs/gridded_pert_scale_{period_number}.nc`. Here, I follow the same pattern as with the `ScaleFactors.nc file`, where we copy an active gridded perturbation file to the base run directory where the Jacobian simulations read that particular file.

### DIFF
--- a/src/components/hemco_prior_emis_component/hemco_prior_emis.sh
+++ b/src/components/hemco_prior_emis_component/hemco_prior_emis.sh
@@ -162,6 +162,7 @@ run_hemco_sa() {
     hemco_start=$1
     hemco_end=$2
     set -e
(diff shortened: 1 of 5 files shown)


msulprizio (approved):
Thanks for these fixes @laestrada!

PR #418, opened 2026-05-19
Fix clustering threshold bugs producing uneven clusters 

iannis-b:
### Name and Institution

Name: Ioannis Binietoglou
Institution: Terra Arc

### Describe the update

Two bug fixes in the clustering algorithm, which could lead to sub optimal clusters:
1) An indexing bug that led to incorrectly estimating the per-cluster DoFs. 
2) Forcing native-resolution cells distorts the `ClusteringThreshold` and produces uneven clusters. The update recalculates the threshold in every iteration to produce more uniform cluster sizes.

#### 1. Wrong indexing when `sensitivities` is an `xarray.DataArray`

`get_highest_labels_threshold` indexes the `sensitivities` argument with `sensitivities[indices]` 
inside its per-label loop.  When `sensitivities` is an `xarray.DataArray`, this expression is interpreted as DataArray's 
default **orthogonal indexing**, not the numpy-style positional indexing the surrounding code assumes. 
The result is a 2-D selection rather than the intended flat selection, which inflates the 
per-label sensitivity sums and produces incorrect clusters.

Fix: if the input has a `.values` attribute, unwrap it to a plain numpy array before the loop, 
so subsequent indexing behaves as numpy positional indexing.

#### 2. Static initial threshold becomes unrepresentative after the first iteration

When `ClusteringThreshold` is not provided in the config, `update_sv_clusters` sets

```
dofs_threshold = sum(flat_sensi) / desired_num_labels
```

once at the start and reuses it for every clustering iteration. 

However, when a large portion of the domain is occupied by forced native-resolution pixels, 
the leftover budget is unrepresentative of the target DOFS-per-cluster for the remaining slots.

For example, let assume we have 12 cells, each with a sensitivity of 0.1; our target is to create 7 clusters,
keeping two of the cells at native resolution. The current algorithm will roughly do the following:
1) The sensitivities of the two native-resolution cells are assigned to 1.1 by `force_native_res_pixels`.
2) `dofs_threshold` is estimated as `(10*0.1 + 2*1.1) / 7 = 3.2 / 7 ≈ 0.46`.
3) With this threshold, the 10 non-native cells end up grouped as one large cluster and several single-cell clusters to fill the grid up to the target cluster number. 

Fix: Recompute `dofs_threshold` at the end of each iteration as:

```
dofs_threshold = remaining_dofs / remaining_cluster_slots
```

where `remaining_dofs` is the sum of sensitivity still attached to unassigned pixels (`labels == 0`) 
and `remaining_cluster_slots = desired_num_labels - max(labels)`.

In the example above, after the fix, the threshold is recomputed once the two native cells are assigned: `remaining_dofs / remaining_cluster_slots = 1.0 / 5 = 0.2`, avoiding oversized clusters.

### DIFF
--- a/src/components/statevector_component/aggregation.py
+++ b/src/components/statevector_component/aggregation.py
@@ -532,10 +532,6 @@ def update_sv_clusters(config, flat_sensi, orig_sv):
 
     print(f"Target DOFS per cluster (ClusteringThreshold): {dofs_threshold}")
 
-    # max cluster size based on user preferences
-    max_cluster_size = (
-        config["MaxClusterSize"] if "MaxClusterSize" in config.keys() else 64
-    )
     max_cluster_size = get_max_cluster_size(config, flat_sensi, desired_num_labels)
 
     try:
@@ -579,8 +575,6 @@ def update_sv_clusters(config, flat_sensi, orig_sv):
 
     print(f"Reducing to {desired_num_labels} elements")
 
-    # total sensitivities of clusters added to the state vector
-    cluster_sensis = []
     # for each agg_level, cluster the data and assign the n_labels
     # with highest total sensitivity to the new label dataset
     for agg_level in cluster_pairs:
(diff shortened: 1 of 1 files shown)


laestrada (approved):
Looks good, thanks for catching these!

PR #420, opened 2026-05-19
Generalize the IMI observation operator using GOOPy

vedrau:

Name: Vedant Rautela
Institution: MIT

### Describe the update

This PR updates the IMI to use GOOPy to generalize the IMI observation operator which heretofore was hardcoded to be TROPOMI-specific. 

Features
- Generalized observation operator using GOOPy
- Decoupling of jacobian construction and observation operator application
- ~~Automated tests to verify jacobian construction and observation operator functional correctness~~
  - ~~Test verifies that jacobian values differ by at most 0.2% (between the current observation operator and original IMI jacobian construction with the original IMI tropomi operator)~~
  - Moved to separate PR

Top-level files/functions
- satellite_operator.py: `apply_operator`, `goopy_apply_operator`, `superobservations`
- jacobian.py
- ~~test_jacobian.py: temp_inversion_workspace, test_jacobian_end_to_end_baseline~~
  - Moved to separate PR

Almost all code outside of `goopy_apply_operator` and `superobservations` is just copy-pasted from the original imi operator


src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -766,15 +1254,20 @@ def average_satellite_observations(
...
         arr["avkern"][idx]  = np.asarray(cell["avkern"], dtype=np.float32)
         arr["observation_count"][idx] = np.float32(cell["observation_count"])
         arr["lat"][idx] = np.float32(cell["lat"])
         arr["lon"][idx] = np.float32(cell["lon"])
+        arr["layer"][idx] = np.arange(n_layers)
 
     return arr
 
+# TODO: update this to add layers to the returned dicts to match what average_sat_observations does 
 def average_satellite_observations_to_CSgrid(
```
vedrau, line 1279: Any guidance on how to test out GOOPy with useGCHP? I've never actually run an inversion with that option before.

And generally, how should I go about testing some of the edge case inversion options?

mcdon1718-h: As far as I know, the GCHP option isn't fully debugged yet; you can try running it but a successful run is not guaranteed - though it would be interesting to see what happens. As for testing edge cases, we don't yet have a formalized testing protocol, so you can either manually test them all yourself, or not test them and include a note in the version update that these options are yet-untested.

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -111,51 +333,346 @@ def apply_average_satellite_operator(
...
+            n_elements,
+            obs_mapped_to_gc,
+            virtual_satellite,
+            config,
+        )
+    else:
+        raise ValueError(f"Error: invalid operator selected: {operator}")
+
+
+def format_goopy_average_satellite_output(
```
vedrau, line 478: Both this function and format_goopy_satellite_output are just helper functions that make sure the output is structured the same way as the original IMI observation operator. There shouldn't actually be any new code in here, all copy-pasted

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -111,51 +333,346 @@ def apply_average_satellite_operator(
...
+    output = {}
+
+    # Always return the coincident satellite and GEOS-Chem data
+    output["obs_GC"] = obs_GC
+    output["GC_index"] = GC_index
+
+    return output
+
+
+def format_goopy_satellite_output(
```
vedrau, line 571: Same comment as above for format_goopy_average_satellite_output

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -269,17 +741,15 @@ def apply_average_satellite_operator(
...
     output["GC_index"] = GC_index
 
-    # Optionally return the Jacobian
-    if build_jacobian:
-        output["K"] = jacobian_K
-
     return output
 
 
+# DEPRECATED: in favor of goopy_apply_operator()
```
msulprizio, line 747: If this is deprecated, is there any reason to leave it here? My preference would be to delete it to avoid confusion. Past code can always be viewed in Git.

src/inversion_scripts/operators/satellite_operator.py
msulprizio: GCHP output within the IMI is in the same format as GCClassic so theoretically it should work. I still need to bring in pull request https://github.com/geoschem/integrated_methane_inversion/pull/379 to get GCHP properly working in the latest version of the IMI. Once both this PR and that are in place, we can fully test GCHP in the IMI and make a separate PR with fixes.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "GOOPy"]
```
msulprizio, line 1: We will need to update ReadTheDocs to instruct users to update git submodules after [downloading the IMI](https://imi.readthedocs.io/en/latest/user-guide/download-code.html) source code.

msulprizio (changes requested):
These changes look good. I only have minor comments. I'm testing on my end now and will let you know if I run into any issues

msulprizio (commented):
I also noticed in commit https://github.com/geoschem/integrated_methane_inversion/pull/420/changes/fa80ccf03f80bfe4beaa1a6e3cd45aea8438fd13 you removed the output files which is what we discussed. Thanks! It's important to note though that these files will remain in the Git history and will therefore inflate the repository size. Could you look into wiping the files from the Git history following the [instructions to remove files from a repository history](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#removing-files-from-a-repositorys-history)?

The other alternative is creating a brand new branch and pull request without those files. You can decide what works best for you.

src/inversion_scripts/operators/satellite_operator.py
vedrau: It is deprecated, but I thought it made sense to leave it in there as an option for people who want to use the original IMI observation operator. These functions are still called in `apply_original_imi_operator` which is routed to if users set `use_goopy=False`

msulprizio: Sounds good. Let's keep it as is for now and have the two options working alongside each other for testing purposes as you originally suggested. We can remove the existing operator in a later pull request.

vedrau:
Closing this PR in favor of [#440](https://github.com/geoschem/integrated_methane_inversion/pull/440) which duplicates this one but without the test files in its history

PR #422, opened 2026-05-19
Kalman period autoadvance

mezger6:

Institution: Terra Arc

### Describe the update

When an interruption occurs during a non-Kalman run, restarting with setup disabled is enough to mostly skip over already completed costly jacobian runs and continue without much overhead and without redoing too much work.

A Kalman run that is interrupted after the first period is completed, however, will restart from the first period, causing very large  amounts of overhead (depending on the number of periods already completed), rerunning already completed jacobians, rerunning (and changing) posteriors and more, making any kind of recovery impossible.

The only way to correctly continue the interrupted run is to use the `FirstPeriod` config variable, but that requires constant supervision and intervention, and would be unfeasible for runs with large numbers of periods.

This PR adds an `AutoAdvanceFirstPeriod` boolean config variable (optional, default `False`) that will autoincrement the `FirstPeriod` config variable in the run config file after each period is successfully completed.

The increment logic is simple - if `kalman.sh`s `run_period()` reaches the end, and the expected files are present in the period folder, the period completed successfully, and it is therefore safe to advance `FirstPeriod`.

This would enable feasible restart automation for Kalman runs, to mitigate spot instance stoppages.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -93,6 +93,7 @@ UpdateFreqDays: 7
 NudgeFactor: 0.1
 MakePeriodsCSV: true
 CustomPeriodsCSV: "/path/to/periods.csv"
+AutoAdvanceFirstPeriod: false
 
 ## State vector
 CreateAutomaticRectilinearStateVectorFile: true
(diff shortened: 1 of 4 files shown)


laestrada (approved):
This looks good, thanks @mezger6!

PR #363, opened 2025-08-10
To implement GCHP into IMI

1Dandan:
### Name and Institution (Required)

Name: Dandan Zhang
Institution: Harvard University

### Describe the update

To implement uniform GCHP and stretched GCHP into IMI.

**IMPORTANT:** This PR also updates the `imi_env.yml` file and will require users to update their local python environments. This can be done using:

```
conda activate imi_env
conda env update -f envs/Harvard-Cannon/imi_env.yml --prune
```

 All changes for IMI scripts are based on IMI v2.2.0, with the changes of the following as the pre-requirement for the IMI-GCHP.
- https://github.com/geoschem/integrated_methane_inversion/pull/355,
- https://github.com/geoschem/integrated_methane_inversion/pull/356
- https://github.com/geoschem/integrated_methane_inversion/pull/365 

Related changes for GCHP source code needed for:
- https://github.com/geoschem/geos-chem/pull/2984
- https://github.com/geoschem/geos-chem/pull/2893
- https://github.com/geoschem/HEMCO/pull/330/.

PR #423, opened 2026-05-19
Miscellaneous fixes

mezger6:

Institution: Terra Arc

### Describe the update

Some minor fixes:

- [‎src/components/jacobian_component/make_jacobian_icbc.py](https://github.com/geoschem/integrated_methane_inversion/commit/6492a52006ef81b1437efb388cbbbdd0d50d7721#diff-ef9ddf97f9a63d2164fbdae7e59409cb98e99adfb918c2aed4f46cabc1c55e01) had a `new_restart.to_netcdf(write_path)` that would fail if the file (but not the dir) is read-only. Changing to an atomic write for safety.
 
- [src/components/kalman_component/kalman.sh‎](https://github.com/geoschem/integrated_methane_inversion/commit/2a4e8301166c8e09dc6b03c696c573cbbbbf9cdb#diff-c663dcfe829ee3ab65e17ade5c2800fb498dc57ab2f622957cd561f3d4c89a05) had some stale `1ppb` that were not changed to `lowbg`

- [‎run_imi.sh](https://github.com/geoschem/integrated_methane_inversion/commit/29d77948b613a9f7a157b804cca2e72d1d5bea93#diff-b30cbde6bcf3696571836af94df19f568bbc27e5e8e596484e7604e68a2e56e6) was calling `submit_job` with a missing parameter (`SchedulerPartition`) to download the TROPOMI data
 
- [‎src/inversion_scripts/plumes.py](https://github.com/geoschem/integrated_methane_inversion/commit/f871a4cb59807a13e6bd4e704ef8ef0538d7666e#diff-56c2087bd6dd4a21320a9fa6c9df835f138101eaa73b33b37456497e75099d5c) had a `d_list.append(ds_i)` outside of a null check for `ds_i` and a `d_names.append(d.myname)` *inside* the check, which would cause a breaking imbalance if the datasource is actually None.


msulprizio (approved):
Thank you for these fixes @mezger6. I will bring them into the `dev` branch and test locally. We should be able to include these in the next release (IMI 2.3.1).

mezger6:
Sorry for the additional laundry list of fixes, I thought it would be better to add them in the "misc" PR instead of opening another one:

- [src/utilities/common.sh](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/97f28fb7029d546fe5d899e78cb1246096ed0a22#diff-c27fc9306ae00cfaf2969ef698d5d0505235f8261b3abb5594b27e06bfea3d9a) - bash treats string beginning with '08' as octal when doing arithmetic, causing trouble for August and the 8th day of the month :)

- [src/components/kalman_component/print_posterior_emissions.py‎](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/526591f52928a397a9c22400a72806041776747a#diff-22464c4088b4bc7022b20d686252c80fec666c1d14cab9c06dc9c294656605a9) used `original_emis_ds` but it is undefined in that file, replaced with the equivalent `hemco_diags` that is set a few lines above.

- similar to the above, [src/components/kalman_component/multiply_posteriors.py](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/36d9e03ace3c27375f865f6526ba48606ec8801b#diff-ab6426e2c2dc40f78ad8ae226accfd8eb4dc713a0255eb957435b867e7019819) referenced `config` but was undefined. Replaced the config value check with a direct check of the actual coordinate names. Unless the names change at some point, this should be quite safe.

- the ReDoJacobian logic in `run_jacobian_simulations.sh` was set to the global final EndDate of the simulation, causing every period except the last one to unnecessarily redo already finished jacobians if restarted. Added a replace step in [src/components/kalman_component/kalman.sh‎](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/6ff70ed5750d219a3e827fffd7fe308b7b44c10f#diff-c663dcfe829ee3ab65e17ade5c2800fb498dc57ab2f622957cd561f3d4c89a05) to set each period's EndDate for each script.

- [src/notebooks/visualization_notebook.ipynb](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/280552db94d3b42e9af7377a5a05134c664b6701#diff-3bfbfafe7520b320cafb3ae5e5ed2401354ca2c6405551455c19f83ffabaa578): replaced a hardcoded './output/netCDF/' with the already set and mkdir'ed `netcdf_path` (was a problem for kalman runs)

- [src/notebooks/kf_notebook.ipynb‎](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/ba936f5abfacd49d66cad30cfa6b0e90f759dc49#diff-62bba7834bfd17c70b020889bc7cbd38efb693d429ce0ec824a9aa714ad39c8b) fixes:
  - missing comma
  - period extraction for `results_prefixes` used all found dirs in `kf_inversions/`, but a non-period `output` dir was throwing it off. Added a regex to only match `period* `dirs.
  - `sf_paths` was set with the full period set read by periods.csv, which would cause inconsistency in a partial run. Replaced with the found dirs we already got for `results_prefixes` for consistency.
  -  a call to `get_posterior_emissions`  was missing the `species` and `OptimizeSoil` params, added them.
  -  add `period1` to path, in order for the imported period dir `utils.py` to be able to import from `classify_TROPOMI_obs_to_CSgrids`
  - add a file check to find dirs with completed periods, and ensure only completed periods are processed (allows running with partially completed or in-progress runs)

- [‎src/inversion_scripts/plumes.py](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/034fcc4db701e141081a4377bbfdbcc77ef8b1e1#diff-56c2087bd6dd4a21320a9fa6c9df835f138101eaa73b33b37456497e75099d5c) had a GHCP related regression where a 2D map (`valid`) was used on 1D GCC coordinates, fixed with a conditional. Also changed a boolean `&` to `and`, could cause issues.

- [src/components/statevector_component/aggregation.py](https://github.com/geoschem/integrated_methane_inversion/pull/423/changes/8f9f6e90943b0a3eafa30c86b91b2a7b13e7d983#diff-1ba17c7dca5338d3eff3eb0ae21199690a629e2818e3239d65b5f79de0343cb0) called `map_sensitivities_to_sv` with an older signature that was changed in PR #363, converted the parameter and the return type to match.

msulprizio (approved):

PR #425, opened 2026-05-26
Update treatment of buffer zone following Nesser et al. (2025)

hannahnesser:

Name: Hannah Nesser
Institution: JPL

### Describe the update

This pull request changes the default treatment of the inversion buffer elements following [Nesser et al. (2025)](https://gmd.copernicus.org/articles/18/9279/2025/). The request removes BufferDeg and nBufferClusters and replaces them with BufferRings and BufferReductionFactor. 

BufferRings is an integer that defines the number of rings of native resolution grid cells to pad the inversion domain with (in addition to the 4 rings used to exclude non-advected grid cells). 

BufferReductionFactor is a float that defines by what factor to reduce the number of native resolution grid cells in the buffer region by K-means aggregation. For example, if there are 100 native resolution grid cells in the buffer region and BufferReductionFactor is 4 (the default value), the clustering algorithm should produce 25 clusters of buffer grid cells. While preserving native resolution in the buffer elements is preferable from the perspective of inversion performance, this is often computationally intractable. BufferReductionFactor is meant to balance computational and quality performance.

There are a few other small changes that were implemented in the NASA-Pleaides environment to enable testing.


msulprizio (approved):

msulprizio:
Thanks again for these updates @hannahnesser. I have confirmed that they work on my end. They produce small changes to the IMI results which I will document here for future reference. These results are from a 1-week run over the permian:

**Reference run (2.3.0):**

```
State vector elements: 175
Prior      emissions : 0.4133052799999029 Tg/y
Posterior  emissions : 0.5033753999597784 Tg/y (mean xhat)
DOFS                 : 0.2261413614058514
```
scale_factors
prior_and_posterior_-_satellite_on_gc_grid

**Dev run (with BC updates):**

```
State vector elements: 100
Prior      emissions : 0.4133052799999029 Tg/y
Posterior  emissions : 0.46623476656371554 Tg/y (mean xhat)
DOFS                 : 0.1897592966211561
```
scale_factors
prior_and_posterior_-_satellite_on_gc_grid

src/utilities/make_state_vector_file.py
```diff
@@ -75,6 +69,9 @@ def cluster_buffer_elements(data, num_clusters, offset):
     # labels = np.zeros(Z.shape)
     valid_indices = np.where(Z == 0)[0]
 
+    # Get the number of clusters
+    num_clusters = int(np.floor( len(valid_indices) / 4 ))
```
laestrada, line 73: @hannahnesser is the 4 here supposed to be config["BufferReductionFactor"]? I don't see where that variable is used anywhere in the code?

PR #442, opened 2026-07-22
Update read satellite function to have callers handle None returns

vedrau:

Name: Vedant Rautela
Institution: MIT

### Describe the update

Update read satellite function to have callers handle None return

### DIFF
--- a/src/inversion_scripts/utils.py
+++ b/src/inversion_scripts/utils.py
@@ -4,6 +4,7 @@ import os
 import sys
 import pickle
 from datetime import datetime, timedelta
+from typing import Optional
 import numpy as np
 import xarray as xr
 import cartopy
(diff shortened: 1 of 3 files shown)


src/inversion_scripts/utils.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import pickle
 from datetime import datetime, timedelta
+from typing import Optional
```
mcdon1718-h: does this require an update to any imi env files? it doesn't seem to be in any of the Harvard cannon envs.

vedrau: No the typing library is built-in and doesn't need to be installed. Also I learned that `Optional` is legacy syntax so I updated it to use the `| None` syntax to specify that the function can return `None`

msulprizio (approved):

PR #447, opened 2026-08-05
Fix `UseWaterObs` always evaluating to True in `jacobian.py`

laestrada:

Name: Lucas Estrada
Institution: Harvard ACMG

### Describe the update
Fix `UseWaterObs` always evaluating to True in `jacobian.py` and use the same pattern for other boolean cli args.

### Problem

`jacobian.py` receives several booleans as positional command-line arguments from the driver shell scripts (`run_inversion.sh`, `posterior.sh`), which pass them as the literal strings `"True"` / `"False"`. In `jacobian.py` these were stored as raw strings and then passed straight through into `params["use_water_obs"]`, which eventually reaches `filter_tropomi` / `filter_blended` in `src/inversion_scripts/utils.py` as a plain `if use_water_obs:` check.

Because a non-empty string is always truthy in Python, `use_water_obs` was effectively`True` for every run, regardless of the `UseWaterObs` setting in the config file. Water observations were therefore retained in the observation filter even when the user had explicitly disabled them.

I also adjusted the other booleans `build_jacobian`, `isPost`, and `viz_prior` to use the same pattern to avoid future confusion.
```python
use_water_obs = sys.argv[13].strip().lower() == "true"
isPost        = sys.argv[14].strip().lower() == "true"
build_jacobian = sys.argv[16].strip().lower() == "true"
viz_prior     = sys.argv[17].strip().lower() == "true"
```

### DIFF
--- a/src/components/posterior_component/posterior.sh
+++ b/src/components/posterior_component/posterior.sh
@@ -296,7 +296,7 @@ run_posterior() {
     kf_period=1
 
     printf "\n=== Calling jacobian.py to sample posterior simulation (without jacobian sensitivity analysis) ===\n"
-    python ${InversionPath}/src/inversion_scripts/jacobian.py ${InvDir} ${ConfigPath} $StartDate_i $EndDate_i $LonMinInvDomain $LonMaxInvDomain $LatMinInvDomain $LatMaxInvDomain $nElements $Species $satelliteCache $SatelliteProduct $UseWaterObs $isPost $kf_period $buildJacobian False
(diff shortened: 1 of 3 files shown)


vedrau (approved):

vedrau:
Idk if this will actually cause any problems but I believe that when we call `config = load_config(...)` the returned `config` dict actually stores the values as booleans i.e. `config = {"UseWaterObs": True, ...}` instead of `config = {"UseWaterObs": "true", ...}`

laestrada:
@vedrau It would have worked because we werent actually using the loaded dictionary config["UseWaterObs"]. Instead we were using a passed-in argument. 

However, we should be using config to load all non-runtime derived arguments when we can to avoid bugs like this and future confusion. I have updated the PR to remove all non-runtime derived arguments and load the variables from the config -- thanks for the improvement idea!

msulprizio (approved):

PR #449, opened 2026-08-06
Add UseGOOPy config setting

vedrau:


mcdon1718-h (approved):
Looks good to me!

msulprizio (commented):
This update looks good. If we're planning on keeping this as an option for a release version then we should add this logical to `config.yml` (and example ones in `envs/`) and to documentation. If it's just here for testing temporarily, then we can leave it as-is.

msulprizio (approved):

vedrau:
> This update looks good. If we're planning on keeping this as an option for a release version then we should add this logical to `config.yml` (and example ones in `envs/`) and to documentation. If it's just here for testing temporarily, then we can leave it as-is.

Yeah I think we want to add it in the next release so that users have the option to turn it off if they're running something that isn't compatible with GOOPy yet (i.e. GCHP)

PR #455, opened 2026-08-12
Update satellite_operator.py to not create .nc_superobservations directories in the satellite data directory

vedrau:

Name: Vedant Rautela
Institution: MIT

### Describe the update

The GOOPy code before this PR was accidentally creating and storing superobservations files/directories where the tropomi/satellite data lived (because of a quirk in `os.path.join` and absolute file paths). Now the superobservations live in the output directory in the `superobservations` subdirectory


vedrau:
We should probably squash and merge this stack of PRs (lots of rebasing and merging that might be confusing to have in the commit history)

msulprizio (approved):

PR #457, opened 2026-08-20
Generalize the IMI to arbitrary satellite products (including MethaneSAT)

vedrau:

Name: Vedant Rautela
Institution: MIT

### For PR Reviewers
Main files of interest:
- satellite_products (the whole directory) --> this is where satellite-product-specific behavior should live
  - products.py is where the meat is
- superobservation.py --> everything involving the creation of standardized superobservations
  - Worth reviewing to see what we assume what data a satellite file has in order to create superobservations  
- msat_funcs.py --> everything specific to MethaneSAT
  - average_methanesat_observations is the top-level function

Notes:
- The test files are all vibe-coded and I think they're _okay_ at the moment (but not worth spending too much time looking at). 
- I edited jacobian.sh to fix an error with a sed command silently failing when using a custom state vector file. It's not really related to the other changes but I'm including it here anyway. 

### Describe the update

This PR expands the IMI’s satellite support beyond TROPOMI, adds end-to-end MethaneSAT support, and introduces a product-neutral superobservation workflow shared by all satellite products. It also improves validation, grid handling, error reporting, and extensibility for future satellite data sources.

### Major changes
- Created the SatelliteProduct class. Each satellite product has product-specific:
  - Raw parsing and filtering
  - Input-directory validation
  - Superobservation generation
  - Preview formatting
  - GOOPy raw-product name
  - Visualization policy
  - Automatic download behavior
  - Datetime parsing
- Added a satellite-product registry with strategies for:
  - TROPOMI
  - Blended TROPOMI 
  - MethaneSAT
- Product-neutral superobservations
- Added a canonical NetCDF schema for IMI superobservations
- Added a standard GOOPy parser for canonical superobservation files
- Preserved product-specific parsing and averaging before conversion to the canonical format

To add a new satellite product, you need to create a new SatelliteProduct and all of its required functions (including a GOOPy parser). The SatelliteProduct classes replace scattered if satellite_product == ... branches and provides a single extension point for future satellite products.

#### MethaneSAT support
MethaneSAT is now supported in the IMI. The IMI creates superobservations by averaging large MethaneSAT L3 files onto the GEOS-Chem grid and assuming a fixed reference averaging kernel and prior profile

#### Preview fixes
- Miscellaneous fixes
  - Standardized the product preview contract so timestamps must be datetime-like.
  - Converted MSAT’s internal YYYYMMDD_HH timestamps before returning preview data.
  - Added shared datetime normalization at the preview boundary.

#### Jacobian setup robustness
- Updated Jacobian HEMCO configuration editing to support custom state-vector locations.
- Made the sed replacement robust to active or commented scale-element lines.
- Added checks that fail immediately when the expected HEMCO entries are not updated.
- Prevented silent configuration failures that previously surfaced later as missing HEMCO scale IDs.

#### Known limitations
- Boundary-condition generation remains TROPOMI-specific and was intentionally excluded from the satellite-product refactor
- MSAT currently uses a fixed averaging kernel and prior profile.
- MSAT superobservation generation currently supports rectilinear GEOS-Chem grids, not GCHP
- use_water_obs has no effect for MSAT because the available fields do not provide a surface classification.
- GOOPy still reads configuration from sys.argv during import; removing this import-time behavior is recommended separately.

#### Testing
Vibe-coded some tests. They aren't that useful at present. Will need to beef this up eventually.


src/utilities/validate_observation_directory.py
```diff
@@ -1,35 +1,66 @@
...
 
-    number_of_unique_orbit_numbers = len(
-        set(all_orbit_numbers)
-    )  # forming a set drops the duplicates
-    number_of_files_in_Sat_datadir = len(files)
 
-    assert (
-        number_of_unique_orbit_numbers == number_of_files_in_Sat_datadir
-    ), "Duplicate orbit numbers found in TROPOMI datafiles. Remove duplicate files from system before continuing."
+def validate_observation_directory(observation_dir, satellite_product):
```
vedrau, line 19: Update this to not silently pass for new satellite products

src/utilities/test_TROPOMI_dir.py
```diff
@@ -1,35 +1,66 @@
...
-        number_of_unique_orbit_numbers == number_of_files_in_Sat_datadir
-    ), "Duplicate orbit numbers found in TROPOMI datafiles. Remove duplicate files from system before continuing."
+def validate_observation_directory(observation_dir, satellite_product):
+    files = list(Path(observation_dir).glob("*.nc"))
+    if not files:
+        raise ValueError(f"No NetCDF observation files found in {observation_dir}")
+    if satellite_product in {"TROPOMI", "BlendedTROPOMI"}:
+        check_for_duplicate_orbit_numbers(observation_dir)
+    elif satellite_product == "MSAT":
+        check_methanesat_files(observation_dir)
```
vedrau: This validation check doesn't actually do anything. Remove it and pass

src/utilities/test_TROPOMI_dir.py
```diff
@@ -1,35 +1,66 @@
...
-    Takes advatnage of the fact that there should be the same number of unique orbit numbers as there are files
-    """
 
-    files = sorted(glob.glob(Sat_datadir + "/*.nc"))
+def check_methanesat_files(observation_dir):
+    """Check filenames and the minimal L3 variables used by the MSAT averager."""
+    files = sorted(Path(observation_dir).glob("*.nc"))
+    filenames = [path.name for path in files]
+    if len(set(filenames)) != len(filenames):
+        raise ValueError("Duplicate MethaneSAT filenames found")
```
vedrau: These lines don't actually do anything useful (duplicate filenames aren't possible on Linux filesystem). Get rid of them

src/utilities/test_TROPOMI_dir.py
vedrau: Actually validating that the files aren't missing variables is maybe useful. Consider keeping

src/utilities/validate_observation_directory.py
```diff
@@ -1,35 +1,66 @@
...
-import glob
 import sys
+from pathlib import Path
+
+import netCDF4 as nc
+
+from src.inversion_scripts.utils import extract_observation_date
+
 
+def check_for_duplicate_orbit_numbers(observation_dir):
```
vedrau, line 9: Revert this function back to the original check_for_duplicate_orbit_numbers. No need to change it here

src/inversion_scripts/utils.py
```diff
@@ -895,6 +910,35 @@ def filter_blended(blended_data, xlim, ylim, startdate, enddate, use_water_obs=F
         return np.where(valid_idx & (blended_data["surface_classification"] != 1))
 
 
+# TODO: rn use_water_obs doesn't have any effect on this function
+def filter_MSAT(data, xlim, ylim, startdate, enddate, use_water_obs=False):
+    """
+    Description:
```
vedrau: Update description

src/inversion_scripts/utils.py
```diff
@@ -895,6 +910,35 @@ def filter_blended(blended_data, xlim, ylim, startdate, enddate, use_water_obs=F
         return np.where(valid_idx & (blended_data["surface_classification"] != 1))
 
 
+# TODO: rn use_water_obs doesn't have any effect on this function
+def filter_MSAT(data, xlim, ylim, startdate, enddate, use_water_obs=False):
```
vedrau: Add type hints

src/inversion_scripts/utils.py
```diff
@@ -895,6 +910,35 @@ def filter_blended(blended_data, xlim, ylim, startdate, enddate, use_water_obs=F
...
+        & (data["latitude"] > ylim[0])
+        & (data["latitude"] < ylim[1])
+        & (data["time"] >= startdate)
+        & (data["time"] <= enddate)
+        & (data["longitude_bounds"].ptp(axis=2) < 100)
+        & (data["latitude"] > -60)
+    )
+
+    return np.where(valid_idx)
+    # if use_water_obs:
```
vedrau: MSAT has some qa flags but I don't think any of them is a surface classification. Maybe we just delete this bit about water obs and say in the description that the use_water_obs has no effect

src/inversion_scripts/utils.py
```diff
@@ -23,6 +24,20 @@
 from pathlib import Path
 
 
+def extract_observation_date(filename: str | Path) -> datetime:
```
vedrau: Make this function have one regex per satellite product (and make it clearer which regex corresponds to which satellite product). It's a bit confusing that right now it just tries two different regexes to get an acquisition time

src/inversion_scripts/jacobian.py
```diff
@@ -273,26 +271,32 @@ def process(filename):
...
-                    "use_water_obs": use_water_obs,
-                },
-                obs_mapped_to_gc,
-                config,
-            )
+            # The MSAT L3 grid can contain tens of millions of source pixels.
+            # Running a second, unaveraged operator solely for visualization is
+            # prohibitively expensive; visualize the canonical superobservations
+            # instead. Other products retain the legacy unaveraged output.
+            if satellite_product == "MSAT":
```
vedrau: Change this to check each satellite product (if there's an unrecognized satellite product, it shouldn't silently fall into the else branch)

src/inversion_scripts/imi_preview.py
```diff
@@ -69,7 +73,27 @@ def get_satellite_data(
...
+        )
+        satellite_data["lat"] = observations["lat_sat"].tolist()
+        satellite_data["lon"] = observations["lon_sat"].tolist()
+        satellite_data[species] = observations[species].tolist()
+        satellite_data["swir_albedo"] = [np.nan] * len(observations)
+        satellite_data["time"] = observations["time"].tolist()
+        satellite_data["observation_count"] = observations[
+            "observation_count"
+        ].tolist()
+        return satellite_data
```
vedrau: Add elif/else branch here for other satellite products

src/inversion_scripts/operators/superobservation.py
```diff
@@ -0,0 +1,152 @@
...
+        "observation_count": "1",
+    }
+    for variable, unit in units.items():
+        dataset[variable].attrs["units"] = unit
+
+    validate_superobservation_dataset(dataset)
+    return dataset
+
+
+def validate_superobservation_dataset(dataset):
```
vedrau: Do we need this validation function? We only call it right above after we've constructed the dataset to our specification

src/inversion_scripts/operators/superobservation.py
```diff
@@ -0,0 +1,152 @@
...
+    pressure_weight = dry_air / dry_air_total
+
+    times = pd.to_datetime(
+        observations["time"], format="%Y%m%d_%H"
+    ).to_numpy(dtype="datetime64[ns]")
+
+    field_names = set(observations.dtype.names)
+    latitude = observations["lat"] if "lat" in field_names else observations["lat_sat"]
+    longitude = observations["lon"] if "lon" in field_names else observations["lon_sat"]
+    data_vars = {
```
vedrau, line 105: Must all of these data vars be included in here? Is it possible that another file format will be missing any of these variables and they should be optional?

src/inversion_scripts/operators/superobservation.py
```diff
@@ -0,0 +1,152 @@
...
+    dry_air_total = dry_air.sum(axis=1, keepdims=True)
+    if np.any(~np.isfinite(dry_air_total)) or np.any(dry_air_total <= 0):
+        raise ValueError("Every superobservation must have a positive dry-air column")
+
+    # The legacy IMI structure stores the prior as a partial column (mol m-2).
+    prior_profile = np.divide(
+        np.asarray(observations["apriori"], dtype=np.float64),
+        dry_air,
+        out=np.full_like(dry_air, np.nan),
+        where=dry_air > 0,
```
vedrau: Shouldn't every dry air subcolumn be positive? Can't we assert this above and then just do regular element-wise division?

As-is, this returns an array with each element being either prior/dry_air or NaN

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -425,7 +377,10 @@ def goopy_apply_operator(
     temp_config.close()
     
     # Set the config file path for GOOPy and import
-    sys.argv[1] = temp_config.name
+    if len(sys.argv) > 1:
```
vedrau, line 328: When would we have more than one arg? When we call GOOPy and pass a custom config file?

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -400,12 +346,18 @@ def goopy_apply_operator(
...
-        'BlendedTROPOMI': 'TROPOMI_blended',
-        'TROPOMI': 'TROPOMI',
-    }
-    goopy_satellite_name = satellite_product_to_goopy_name_map.get(satellite_product, satellite_product)
+    if operator == "satellite_average":
+        # All averaged products are persisted using the same product-neutral
+        # schema, irrespective of the source retrieval product.
+        goopy_satellite_name = "IMI_superobservation"
+    else:
+        satellite_product_to_goopy_name_map = {
```
vedrau: Update the GOOPy config so that the names match and we don't need this mapping

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -43,7 +51,9 @@ def get_goopy_config_path():
         "or the repository root. Make sure the IMI repository root is on PYTHONPATH."
     )
 
-def obs_to_xarray_dataset(obs_mapped_to_gc, species, filename, config):
+def obs_to_xarray_dataset(
```
vedrau: Type hints

vedrau: Actually just get rid of this function, it literally just calls the other one

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -296,52 +216,78 @@ def superobservations(
...
+            state_vector_path,
+            species=species,
+            time_threshold=time_threshold,
+            gc_startdate=gc_startdate,
+            gc_enddate=gc_enddate,
         )
-        GC_shape = (len(gc_lat_lon['lat']), len(gc_lat_lon['lon']))
+        gc_lat_lon = get_gc_lat_lon(gc_cache, gc_startdate)
+        GC_shape = (len(gc_lat_lon["lat"]), len(gc_lat_lon["lon"]))
+    else:
```
vedrau: Add elif/else branch for tropomi. New satellite product should fail

src/inversion_scripts/operators/operator_utilities.py
```diff
@@ -141,6 +142,7 @@ def get_gridcell_list(lons, lats, species):
                     "lon_sat": [],
                     "observation_count": 0,
                     "observation_weights": [],
+                    # NOTE: this is the number of vertical levels in the GC grid
```
vedrau: I don't think this is true. Also what is this magic number for layers? Define a constant

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
+from netCDF4 import Dataset
```
vedrau: Generally clean up this file

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+import pandas as pd
+import xarray as xr
+from src.inversion_scripts.utils import get_strdate
+
+
+# def make_custom_bc_files():
+
+
+
+def adjust_column_average_concentration_time_avg(met_file, conc_file, target_column_avg, output_file):
```
vedrau: I think this function isn't used anywhere

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+            print(f"Processing date: {current_date.strftime('%Y-%m-%d')}")
+            adjust_column_average_concentration_time_avg(met_file, conc_file, target_column_avg, output_file)
+        else:
+            print(f"Skipping date {current_date.strftime('%Y-%m-%d')}: Input files not found.")
+        
+        # Move to the next date
+        current_date += datetime.timedelta(days=1)
+
+
+def get_sv_ll_bnds(state_vector):
```
vedrau: Type hints + description

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+    sv = xr.open_dataset(state_vector)
+    lat_bnds = np.zeros(len(sv['lat']+1))
+    lon_bnds = np.zeros(len(sv['lon']+1))
+    
+        
+    lat_inc = (sv['lat'][1]-sv['lat'][0])/2
+    lon_inc = (sv['lon'][1]-sv['lon'][0])/2
+
+
+    print(lon_inc)
```
vedrau: delete

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+    sv = xr.open_dataset(state_vector)
+
+
+    sv_ll['lon'] = sv['lon']
+    sv_ll['lat'] = sv['lat']
+    
+
+    return sv_ll
+
+def _legacy_average_methanesat_observations(
```
vedrau: delete

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+
+
+def average_methanesat_observations(
+    data_path,
+    state_vector,
+    species="CH4",
+    time_threshold=None,
+    gc_startdate=None,
+    gc_enddate=None,
+    file_id="MSAT",
```
vedrau: delete

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+):
+    """Average a rectilinear MethaneSAT L3 file onto the state-vector grid.
+
+    The input can be much larger than memory. Only coordinate rows intersecting
+    the state-vector domain are read, and the two-dimensional retrieval fields
+    are accumulated in row chunks. A single joint validity mask is applied to
+    XCH4, time, and surface pressure so all averaged fields and counts describe
+    the same contributing pixels.
+    """
+    del file_id  # Retained for API compatibility.
```
vedrau: delete

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+        ("nir_albedo", "f4"), ("swir_albedo", "f4"),
+        ("dry_air_subcolumns", "f4", (19,)),
+        ("apriori", "f4", (19,)), ("avkern", "f4", (19,)),
+        ("layer", "f4", (19,)), ("observation_count", "f4"),
+        ("lat", "f4"), ("lon", "f4"),
+    ]
+    return np.zeros(size, dtype=dtype)
+
+
+def gosat_pressure_grid(psurf,ptrop):
```
vedrau: Type hints + better description

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+    p_strat = [8.0000000e1, 5.0000000e1, 1.0000000e1, 1.0000000, 1.0000000e-1]
+    for l,pl in zip(range(lmx_t+2,lmx+1),p_strat):
+        pedge[:,:,l] = pl
+
+    # Intermediate strat level
+    pedge[:,:,lmx_t+1] = 0.5*(pedge[:,:,lmx_t]+pedge[:, :, lmx_t+2])
+
+    return pedge
+
+def compute_profile_constgrav(pedge, g0=9.80665):
```
vedrau: Type hints + description

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+
+    outp['air_vcd'] = np.zeros_like(outp['pmid'])
+    for iz in range(zmx):
+#        outp['air_vcd'][:,:,iz] = (outp['pedge'][:,:,iz]-outp['pedge'][:,:,iz+1])/g0 * (1/(outp['mw'][:,:,iz]))
+        outp['air_vcd'][:,:,iz] = (outp['pedge'][:,:,iz]-outp['pedge'][:,:,iz+1])*100/g0 * (1/mw_dry)
+    outp['dry_air_vcd'] = outp['air_vcd']#*(1.0-xh2o)
+
+    return outp
+
+def group_data_in_gridbox(data, lat2d, lon2d, lat_bnds, lon_bnds, fill_value=1e30):
```
vedrau: Do we still use this function? If not, delete. Else, type hints

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+
+    if gc_lats.ndim != 1 or gc_lons.ndim != 1:
+        raise ValueError("MethaneSAT averaging requires 1-D state-vector lat/lon")
+    if len(gc_lats) < 2 or len(gc_lons) < 2:
+        raise ValueError("State-vector lat/lon must each contain at least two cells")
+
+    def coordinate_edges(centers):
+        differences = np.diff(centers)
+        if not np.all(differences > 0):
+            raise ValueError("State-vector coordinates must be strictly increasing")
```
vedrau: Is this true? It seems reasonable but idk if we should be erroring out (i.e. do other places in the IMI depend on this assumption?)

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+    with xr.open_dataset(state_vector) as state_vector_ds:
+        gc_lats = np.asarray(state_vector_ds["lat"].values, dtype=np.float64)
+        gc_lons = np.asarray(state_vector_ds["lon"].values, dtype=np.float64)
+
+    if gc_lats.ndim != 1 or gc_lons.ndim != 1:
+        raise ValueError("MethaneSAT averaging requires 1-D state-vector lat/lon")
+    if len(gc_lats) < 2 or len(gc_lons) < 2:
+        raise ValueError("State-vector lat/lon must each contain at least two cells")
+
+    def coordinate_edges(centers):
```
vedrau: This seems to be assuming a uniform resolution state vector, but SV elements can have varying resolution. Rework this averaging function to not depend on uniform res elements

vedrau: Also type hints

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+    )
+
+    sums_xch4 = np.zeros(n_cells, dtype=np.float64)
+    sums_time = np.zeros(n_cells, dtype=np.float64)
+    sums_surface_pressure = np.zeros(n_cells, dtype=np.float64)
+    sums_latitude = np.zeros(n_cells, dtype=np.float64)
+    sums_longitude = np.zeros(n_cells, dtype=np.float64)
+    counts = np.zeros(n_cells, dtype=np.int64)
+
+    def values_with_nan(variable, key):
```
vedrau: Type hints + description

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+        observations["time"][index] = (
+            get_strdate(timestamp, time_threshold)
+            if time_threshold is not None
+            else timestamp.round("60min").strftime("%Y%m%d_%H")
+        )
+
+    return observations
+
+
+def average_methanesat_observations(
```
vedrau: This function is too big; break up into smaller functions with a single purpose

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+        return _empty_methanesat_observations(species)
+
+    mean_xch4 = sums_xch4[populated] / counts[populated]
+    mean_time = sums_time[populated] / counts[populated]
+    mean_surface_pressure = sums_surface_pressure[populated] / counts[populated]
+    mean_latitude = sums_latitude[populated] / counts[populated]
+    mean_longitude = sums_longitude[populated] / counts[populated]
+    j_gc, i_gc = np.unravel_index(np.flatnonzero(populated), (n_lat, n_lon))
+
+    # Fixed profiles retained from the original MethaneSAT implementation.
```
vedrau: IMPORTANT: should we be looking at the apriori data or do we actually want to use fixed reference AK and prior?

src/inversion_scripts/operators/msat_funcs.py
```diff
@@ -0,0 +1,749 @@
...
+            get_strdate(timestamp, time_threshold)
+            if time_threshold is not None
+            else timestamp.round("60min").strftime("%Y%m%d_%H")
+        )
+    return observations
+
+
+def _empty_methanesat_observations(species, size=0):
+    """Allocate the standard IMI representation for 19-layer MSAT data."""
+    dtype = [
```
vedrau: This is the IMI superobservation dtype. I should define this somewhere high-level and then use this as a type-hint and dtype wherever needed

src/inversion_scripts/utils.py
vedrau: Update: maybe bring this function into the SatelliteProduct object

src/inversion_scripts/operators/satellite_operator.py
```diff
@@ -596,7 +522,7 @@ def format_goopy_satellite_output(
     """
     satellite, sat_ind = read_and_filter_satellite(
```
mcdon1718-h: this is the only call that's missing None handling

src/inversion_scripts/operators/operator_utilities.py
```diff
@@ -13,6 +13,9 @@
 
 warnings.filterwarnings("ignore", category=UserWarning, module="xarray")
 
+# TROPOMI's 13 retrieval pressure edges bound 12 vertical layers.
+TROPOMI_N_LAYERS = 12
```
msulprizio, line 17: Ideally, we would make this "SATELLITE_N_LAYERS" and not hardcode to TROPOMI.

msulprizio (approved):
These updates look good. We will need to update the IMI documentation for the additional options. That can be done in a separate PR

src/inversion_scripts/operators/operator_utilities.py
vedrau: The way I've formatted these changes the entire `operator_utilities.py` file is TROPOMI-specific (we should probably rename it to `tropomi_utilities.py`). All of the MSAT stuff lives in `msat_funcs.py` and I think we can continue this pattern for additional satellites

msulprizio: Thanks for clarifying @vedrau. That sounds good to me.

PR #460, opened 2026-09-01
Update Ja normalization to divide by DOFS instead of n_elements

vedrau:

Name: Vedant Rautela
Institution: MIT

### Describe the update

Update Ja normalization to divide by DOFS instead of n_elements. This affects the selection of ensemble members which are now selected if they satisfy `0.5 <= J_a/DOFS <= 2`

### DIFF
--- a/src/inversion_scripts/invert.py
+++ b/src/inversion_scripts/invert.py
@@ -508,11 +508,14 @@ def do_inversion(
     # Averaging kernel matrix (use unweighted Sa)
     A = np.identity(n_elements) - S_post @ inv_Sa
 
+    # Degrees of freedom for signal (DOFS)
+    DOFS = A.trace()
+
     # Calculate J_A, where delta_optimized = xhat - xA
     # J_A = (xhat - xA)^T * inv_Sa * (xhat - xA)
     delta_optimizedT = delta_optimized.transpose()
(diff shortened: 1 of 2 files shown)


laestrada (commented):
Looks good, but I think we need a corresponding change in lognormal_invert.py

sarahhancock (approved):
Looks great to me! One thing to be mindful of is that if DOFS is zero, there would be an error, but I feel like this would almost never happen so I don't think we need to do anything about it...
Thanks for implementing this!

msulprizio:
Thanks for this PR @vedrau. I am currently testing on my end with a 1-month Permian inversion.

msulprizio:
The consensus of the IMI Steering Committee is to stick with Ja/n for now.

Daniel Jacob wrote:
> JA/DOFS = 1 works only if the error covariance matrices are correct. So our problem has been that we generally don’t account for prior error correlation. When [Sarah] accounts for a realistic background prior error covariance and observational error covariance she gets the desired result of JA/DOFS ~ 1.  
>
>JA/n  ~ 1 as a diagnostic is not wrong because it does protect against overfit, but it is based on incorrect error covariance matrices. The solution will be to use correct error covariance matrices.  

@sarahhancock is still investigating and will provide an update when mature.

PR #463, opened 2026-09-05
Replace usage of explicit inverse in `delta_optimized` with a direct solve

marcus-campbell:

Name: Marcus Campbell
Institution: Independent

### Describe the update

This changes the calculation of `delta_optimized` to use `np.linalg.solve`. The previous implementation explicitly formed the inverse of the constraint matrix and then multiplied it by the right-hand side.

This expresses the linear solve directly and avoids constructing an unnecessary inverse.

I compared the two implementations with representative inputs on NumPy 1.24.4, 1.26.4, and 2.5.0. The output shapes and dtypes were unchanged. Numerical differences were at floating-point rounding scale; the largest observed relative difference was `8.52e-15`.

### DIFF
--- a/src/inversion_scripts/invert.py
+++ b/src/inversion_scripts/invert.py
@@ -482,8 +482,9 @@ def do_inversion(
     )
 
     # Solve for posterior scale factors xhat using the weighted constraint matrix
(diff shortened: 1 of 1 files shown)


laestrada (approved):
Thanks @marcus-campbell, this is a nice improvement. @msulprizio, .solve() should be equivalent to np.linalg.inv, but is more memory efficient.

PR #464, opened 2026-09-07
Use mean emissions in y-F(x') term for lognormal inversion

### LINKED ISSUE #462: Fix for inversion using lognormal prior error PDF

Name: Melissa Sulprizio
Institution: Harvard

### Description of your issue or question

From Daniel Jacob:

>When using the lognormal inversion with x’=ln(x), the (y-F(x’)) term at each iteration i should be calculated as F(x’i) = Kxi + c  where xi is the mean emission (IMI uses the median): xi = exp[x’i + s’i/2] with s’ being the GSD. This is clearest from the derivation of the Gauss-Newton equation, which is a limiting case of Levenberg-Marquardt.  Sarah finds that this solves her problem of the mean solution consistently over predicting observations, and Xiaolin finds that it makes the OSSE solutions look better.

xlwangg:

Name: Xiaolin Wang
Institution: Harvard

### Describe the update
The lognormal inversion optimizes x' = ln(x), where exp(x') is the median of x. The state vector and the prior term J_A are therefore expressed in terms of the median. However, in the observation term, (y - F(x')) is the mismatch between observed and simulated concentrations, and the simulation is driven by the mean emissions. It is the mean of the emissions, not the median, that we require to be consistent with the observations.

IMI currently evaluates F(x') = K·x_n with x_n the median of x, which biases simulated concentrations low and drives the posterior emissions high. The updated code now evaluates the observation term as y - K·x_mean, with the mean emissions updated at each iteration.

### DIFF
--- a/src/inversion_scripts/lognormal_invert.py
+++ b/src/inversion_scripts/lognormal_invert.py
@@ -155,7 +155,6 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         # the lognormal part of K gets scaled by prior_scale to convert to median
         K_ROI = K_temp[:, :-num_normal_elems]
         K_normal = K_temp[:, -num_normal_elems:]
-        K_ROI = prior_scale * K_ROI
         K_full = np.concatenate((K_ROI, K_normal), axis=1)
 
         m, n = np.shape(K_ROI)
@@ -247,6 +246,12 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         # ln(xn) = x(n-1) + inv(term1+term2)@(term3 + term4)
         # where x(n-1) is the previous iteration of xn until convergence
         print("Status: Iterating to calculate ln(xn)")
+
+        xnmean = np.concatenate(
+            (np.exp(lnxn[:-num_normal_elems]), lnxn[-num_normal_elems:]),
+            axis=0,
+        )   
+    
         while xn_iteration_pct_diff >= convergence_threshold:
 
             # we need to transform lnxn to xn to calculate K_prime
@@ -256,7 +261,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
             )
             # K_prime is the updated jacobian using the new xn from the previous iteration
             K_prime = np.concatenate(
-                (K_ROI * xn[:-num_normal_elems].T, K_normal), axis=1
+                (prior_scale * K_ROI * xn[:-num_normal_elems].T, K_normal), axis=1
             )
 
             # commonly used term for term1 and term3
@@ -267,7 +272,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
             term2 = (1 + kappa) * invlnsa_constraint
             inv_term = np.linalg.inv(term1 + term2)
 
-            term3 = gamma_K_prime_transpose_Soinv @ (y_ybkg_diff - K_full @ xn)
+            term3 = gamma_K_prime_transpose_Soinv @ (y_ybkg_diff - K_full @ xnmean)
             term4 = invlnsa_constraint @ (lnxn - lnxa)
 
             # put it all together to calculate lnxn_update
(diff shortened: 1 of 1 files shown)


msulprizio (approved):
This looks good to me. Thanks @xlwangg! I will await any further comments from @laestrada and @sarahhancock.

src/inversion_scripts/lognormal_invert.py
```diff
@@ -284,29 +289,30 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
 
             lnxn = lnxn_update
 
+            # Calculate averaging kernel and degrees of freedom for signal
+            K_primeT_so = gamma * np.transpose(K_prime) @ Soinv
+            # posterior error covariance matrix (uses unweighted Sa)
+            lns = np.linalg.inv(K_primeT_so @ K_prime + invlnsa)
+
+            # Calculate posterior mean xhat
+            dlns = np.diag(lns[:-num_normal_elems, :-num_normal_elems])
```
laestrada, line 298: Is there a reason why `dlns`, `lns`, and `K_primeT_so` are now included in the iteration? Shouldn't these still be calculated after convergence?

xlwangg: I think it is needed because we want to update shat during the iterations to convert the median to the mean. Does that make sense?

laestrada: Oh yeah, good point -- thanks!

laestrada (approved):
Looks good to me -- thanks Xiaolin!

sarahhancock (commented):
Hi Xiaolin, thanks for doing this! I did this fix over South America too and our code mostly looks the same. I have a couple comments/questions.

We probably talked about this already, so sorry if I forgot or if this is opening another can of worms: should K_prime now be built from xnmean rather than xn? Now that the obs term uses the mean (y − K_full @ xnmean), I think the Jacobian should be at the mean too, but K_prime is currently built from prior_scale * xn (the median). In my code I've been building the Jacobian from the mean. I don't think this would have a large impact so it's probably not a big deal - but I think this comes from the same Gauss–Newton derivation Daniel mentioned.

Let me know if I'm misreading anything and thanks again, this looks great!

xlwangg:
Hi @sarahhancock, thanks for the comment! I could be misunderstanding, but I do build K_prime from the median of xn, since K_prime = d(y)/dln(xn) = xn · d(y)/d(xn), where I interpret xn here as the median. I only use the mean emissions for K_full. What do you think?

sarahhancock:
Hi @xlwangg - right, I was thinking maybe we should be using the mean K instead of the median. My thinking was that since the mismatch already uses the mean (y − K_full @ xnmean), its Jacobian should be the derivative of that same term, which brings down xnmean (the mean), not xn (the median). I think we'd still be optimizing for the median, just making the Jacobian consistent with the mean that's already in the residual. I couldn't remember if we had discussed this already, so feel free to ignore if we already decided it should be the median instead!

xlwangg:
I agree now that we should calculate the Jacobian from the mean emissions. I've made a new [commit](https://github.com/geoschem/integrated_methane_inversion/pull/464/changes/a67573246e237348d304a734d028ff205f2022c9) to fix this. @sarahhancock could you please take another look? Thanks!

sarahhancock:
Hi @xlwangg, it looks great now, thanks again for doing this!

src/inversion_scripts/lognormal_invert.py
```diff
@@ -247,6 +246,12 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
         # ln(xn) = x(n-1) + inv(term1+term2)@(term3 + term4)
         # where x(n-1) is the previous iteration of xn until convergence
         print("Status: Iterating to calculate ln(xn)")
+
+        xnmean = np.concatenate(
```
eastjames: I'm confused here -- `prior_scale` is `1 / np.exp((np.log(float(sa)) ** 2) / 2)`. This is equivalent to $\exp{[-\frac{s'}{2}]}$ in Eq. 5 of Hancock et al. 2025. It converts the mean of a lognormal distribution to the median. So `xnmean` here is actually the median, but I think it should not be. Is that right or am I mixing something up?

src/inversion_scripts/lognormal_invert.py
```diff
@@ -256,7 +261,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
             )
             # K_prime is the updated jacobian using the new xn from the previous iteration
             K_prime = np.concatenate(
-                (K_ROI * xn[:-num_normal_elems].T, K_normal), axis=1
+                (K_ROI * xnmean[:-num_normal_elems].T, K_normal), axis=1
```
eastjames, line 260: Same here, the variable is called mean but I think the value is actually the median.

src/inversion_scripts/lognormal_invert.py
```diff
@@ -267,7 +272,7 @@ def lognormal_invert(config, state_vector_filepath, jacobian_sf):
             term2 = (1 + kappa) * invlnsa_constraint
             inv_term = np.linalg.inv(term1 + term2)
 
-            term3 = gamma_K_prime_transpose_Soinv @ (y_ybkg_diff - K_full @ xn)
+            term3 = gamma_K_prime_transpose_Soinv @ (y_ybkg_diff - K_full @ xnmean)
```
eastjames, line 272: Same as above, variable name does not match its value. Value should be the mean but it is the median I think

eastjames (changes requested):

src/inversion_scripts/lognormal_invert.py
sarahhancock: I think you're right here and this line should drop the prior scale!
I don't think this line would have a big impact on results since it's just the initialization but worth fixing

eastjames: Thanks! I pushed some edits

src/inversion_scripts/lognormal_invert.py
eastjames: I think we don't need to update s_hat with every iteration. I pushed some changes

eastjames:
Hi @xlwangg, thank you for doing this! I pushed a couple edits, could you please take a look and see if you agree?

xlwangg:
Hi @eastjames, thanks for catching this! It looks like there were some conflicts between our edits, so I've pushed a new commit that merges yours in. Could you take another look?

eastjames:
Thanks @xlwangg! That looks good to me, and I see now why we still need to include s in the iteration.

eastjames (approved):

src/inversion_scripts/lognormal_invert.py
eastjames: ^ this was wrong, we do need to update s_hat at each iteration

